PR #69, opened 2020-06-10
Integration tests

ajkiessl:
This test creates a new record in the sandbox every time it runs.  Curious what the limits are on that... guess we'll see.  I can change it to just return an error if we think this will be a problem, but I wanted it to test a successful POST.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
 --format documentation
+--tag ~glacial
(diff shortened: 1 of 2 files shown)


ajkiessl:
@EricDurante that shouldn't be too difficult.  I can just use httparty right there in the test to pull down the data from sandbox, compare it, then I think I would use the "put-code" to identify which record to delete with another post.  I'm not too sure on that last part... need to dig into the documentation a little further.

ajkiessl:
Yeah looks like I would use the "put-code" in the path, but I would use a DELETE not a POST.

ajkiessl:
@EricDurante If it looks good to you, then this is ready to be merged.

EricDurante:
@ajkiessl 
This looks awesome, Alex. Thanks so much for working on this!

PR #74, opened 2020-06-18
this is the closest I've been able to get.  Sopmething is not joining…

ajkiessl:
… properly in the second join to authorships... so it is missing authorships of other users.  May need to link up another set of publications.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,12 +90,14 @@ class User < ApplicationRecord
   end
 
   def self.needs_open_access_notification
-    joins({authorships: :publication}, :user_organization_memberships).
-    where('open_access_notification_sent_at IS NULL OR open_access_notification_sent_at < ?', 6.months.ago).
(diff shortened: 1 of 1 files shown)


ajkiessl:
Hey @EricDurante .  I modified the method to use SQL statements instead.  I'm not too sure this is exactly what we would want in terms of refactoring, but will probably be better in terms of efficiency.  Only problem is I cannot get the second join to authorships to work correctly.  I want the joins to look something like this:

```
users > authorships > publications > authorships > waivers
   V
user_organization_memberships
```
The far right linking of authorships should give us the ability to query on whether or not another authorship (linked to one of the User's publications) already has a waiver or scholarsphere upload.  However, if you run the specs, you'll see they fail in both of these instances.  I've wracked my brains all day yesterday trying to figure it out, but with no success.  If you have the time, could you take a look?  I can revisit this tomorrow possibly, but need to focus my attention on something else today.

PR #75, opened 2020-06-19
Refactor oa notify

EricDurante:
@ajkiessl Sorry, I merged the original branch, so your pull request got closed.

Thanks a lot for taking a shot at this. Incredibly, I've never attempted to do anything quite like this before, and it was definitely **not** easy to wrap my head around what needed to be done. What you contributed helped to get me over the hump.

Anyway, after stretching the absolute limit of my SQL knowledge, I think I got it to work. I don't think that it's possible to do exactly what you had in mind in terms of table joins (I could be wrong), but the statements that I came up with pass the tests and they seem to produce the same results as the old method when run against the production data.

Now that I understand the pattern here, I think I can apply it to a couple of the other queries and maybe refactor some duplication if I get a chance.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,12 +90,14 @@ class User < ApplicationRecord
   end
 
   def self.needs_open_access_notification
-    joins({authorships: :publication}, :user_organization_memberships).
-    where('open_access_notification_sent_at IS NULL OR open_access_notification_sent_at < ?', 6.months.ago).
+    joins({authorships: {publication: :authorships}}, :user_organization_memberships).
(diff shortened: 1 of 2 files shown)


ajkiessl:
@EricDurante Yeah, it's not possible to do it with only joins.  The sub-querying and "IN"/"NOT IN"s is what I needed, and where my knowledge fell short.  I should have dug into the documentation sooner!  Regardless I learned a lot about SQL, so I should have better luck next time.  Thanks for the help and the code looks good to me 👍 .

PR #81, opened 2020-07-20
Orcid pubs integration

ajkiessl:


EricDurante:
Hey @ajkiessl, this is looking great! I'm just going to throw a couple of notes in about questions that I have or things that we might want to look at tomorrow as reminders to myself. No need to respond to anything here in GitHub if you'd rather just discuss in person.

app/controllers/orcid_works_controller.rb
```diff
@@ -0,0 +1,27 @@
+class OrcidWorksController < UserController
+  before_action :authenticate!
+
+  def create
+    authorship = Authorship.find(params[:authorship_id])
```
EricDurante, line 5: Here we're probably going to want to scope this query to the current user's authorships.

app/models/orcid_resource.rb
```diff
@@ -0,0 +1,51 @@
+class OrcidResource
```
EricDurante, line 1: Good abstraction here 👍 👍

app/models/orcid_work.rb
```diff
@@ -0,0 +1,61 @@
...
+        work[:"external-ids"] = { "external-id": [] } unless work[:"external-ids"].present?
+        work[:"external-ids"][:"external-id"] << {
+              "external-id-type": orcid_type.to_s,
+              "external-id-value": (type == :url ? url : publication.send(type).to_s),
+              "external-id-relationship": relationship
+        }
+      end
+    end
+
+    publication.authorships.each do |ext_author|
```
EricDurante, line 35: I'm not certain about this, but here we may want to use the metadata database concept of `Contributor` instead of `Authorship`. That will give us a list of the names of all of the authors on the work rather than just the Penn State-affiliated authors.

app/models/orcid_work.rb
```diff
@@ -0,0 +1,61 @@
...
+      type: 'journal-article',
+      "publication-date": {
+        year: publication.published_on.year,
+        month: publication.published_on.month,
+        day: publication.published_on.day
+      },
+      url: url
+    }
+
+    external_ids.each do |type|
```
EricDurante, line 22: You'll have to explain the orcid "external-id" concept to me...

EricDurante: Oh, never mind. I think I see what's going on. 

It occurs to me that a user of our app won't really be able to see some of the data that we're going to be adding to their orcid works. Some of these bits of data are an example of that.

Knowing that some of this data isn't super clean (especially in the case of things that the user has added to Activity Insight), I wonder if users might be surprised by some of the information that ends up in orcid. That might be a whole additional workflow issue that we may need to consider.

I'm thinking that it might be good to give the user some kind of preview of all of the fields/values that are going to be added to orcid for each work before they confirm that they want to do it. Or at the very least, maybe we need to give users the ability to delete things that they've added to orcid if they see that some of the data isn't quite right after the fact. Actually handling correction of any bad data is a whole other issue, but it would probably be the responsibility of a metadata admin.

app/views/profiles/edit_publications.html.erb
```diff
@@ -47,6 +48,16 @@
...
+          <td>
+            <% if a.user.orcid_access_token && a.publication.orcid_allowed? %>
+              <% if a.orcid_resource_identifier.present? %>
+                <sup>*</sup>This information has been added to your ORCID record.
+              <% else %>
+                <%= button_to "Add to my ORCID Record", orcid_works_path, class: 'btn orcid-button',
+                              remote: false, params: { authorship_id: a.id } %>
+              <% end %>
+            <% end %>
+          </td>
```
EricDurante, line 60: This looks good. As we're adding controls to the UI and/or changing the workflow, we should also just keep in mind that at some point we'll probably want to give users the ability to update and delete works that they've added to orcid.

app/models/publication.rb
```diff
@@ -449,6 +449,10 @@ def has_open_access_information?
     !preferred_open_access_url.blank? || scholarsphere_upload_pending? || open_access_waived?
   end
 
+  def orcid_allowed?
+    doi.present? || url.present? || open_access_url.present?
+  end
```
EricDurante, line 454: Does this reflect a requirement on orcid's end? Or does it just not make sense to create works that don't have any of this information?

In any case, we should take a closer look at these fields to be sure that we're actually getting the data that we want.

ajkiessl: Just a note for our meeting:  From the responses I'm getting from the Orcid API and sifting through their docs, it appears they require at least one "external-id".  However, it gets more complicated than that: that one required external-id must have an "external-id-relationship" of "self".  The value of "external-id-relationship" is a bit arbitrary and up to the client, but in their documentation they don't classify isbns or issns or any other location that does not point you directly to the article as "self"; they are classified as "part-of".  So, what I've gathered from this is that a work must have at least a doi or url as an external-id.  That raises the question: Do we want all of the urls/dois uploaded, or only some?  And which one? (:url, :open_access_url, :user_submitted_open_access_url)

ajkiessl:
@EricDurante  This is ready to be merged when you are ready.  We can discuss when you get back from vacation if you'd like.

EricDurante:
Thanks, Alex!
This looks good. I'll go ahead and merge and deploy it now, and I'll give it a quick test in production.

PR #92, opened 2020-09-23
Export pubs to ai

ajkiessl:
@EricDurante We can discuss then when you get back from break, just some notes in the meantime:

A post request is sent to AI per publication and it includes all users attached to the publication.  Depending on the number of users, each request could take anywhere from 2-10 seconds.  This adds up quickly, so I think we talked about having this run as a background job.  We should be able to do this with ActiveJob.

I've tested the export a bunch, and I've worked out all of the errors.  However, there's always a chance we run into some in the future.  I am logging responses with Rails.logger but this might not be the best way of doing.  Detecting errors may be a little more difficult with large exports like this.

The export also only works assuming every user that has an 'activity_insight_identifier' is in fact in AI.  I guess there may be a chance that someone was removed from AI, and still has the identifier in the RMD.

Finally, I'm not too sure about the interface.  It was difficult working around rails_admin.  I could only get the proper functionality by building off of your index_by_org view, so you can only export to AI if you've indexed by org first.  Every other configuration I tried ended up adding something I didn't want, or not giving me the proper queries.  I can show you this when you get back.


ajkiessl:
Turns out you can't run background jobs with just ActiveJob.  Will need a backend queuing system like SideKiq.

ajkiessl:
Add a check to skip publications with Activity Insight import.  See if we can get IDs to return after exporting to Activity Insight as a response.

ajkiessl:
There's no way to get the IDs of records in the response from Activity Insight.

ajkiessl:
Using delayed_job as the async job handler: https://github.com/collectiveidea/delayed_job/.  It runs as its own process.  To start the process:
`bin/delayed_job start`
to stop the process:
`bin/delayed_job stop`
to restart:
`bin/delayed_job restart`

ajkiessl:
Defaulting all publication "CONTYPES" to "Journal Article".  The only two options we have in AI that would match up to the publication types we allow are: "Journal Article" and "Journal Article, In House".

Defaulting any statuses that are "Accepted/In press" (from Pure) to "Accepted".

ajkiessl:
@EricDurante This is ready to go whenever you are.

ajkiessl:
This is going to need changed if we are pulling in research outputs that are not journal articles.  It also needs conflicts resolved.  Will work on this in the coming week.

EricDurante:
@ajkiessl 
Another possibility for safeguarding against creating duplicate records in Activity Insight occurred to me. We could simply only allow a publication to be exported from RMD to Activity Insight one time.

We could add a boolean field to publications in RMD, `exported_to_activity_insight` or something like that, and whenever we send a publication to Activity Insight and get a success response back, we could set that field to `true`. Then, whenever we're exporting publications to Activity Insight, in addition to checking whether or not each publication was originally imported from AI, we can also look at whether or not it has been exported to AI before.

Just another suggestion. Feel free to disregard if you see any problems with it or if you just don't think that it's necessary.

ajkiessl:
@EricDurante I'm still a little indecisive on this.  This ☝️ would be a simple and effective way of doing onboarding.  However, if we want to have the ability to update records between RMD and AI, setting up AI to have the "RMD ID" as a primary key to update by would give us that ability.  On the other hand, that would add a pretty vital coupling/dependency between the two systems.  If that primary key were to change... then we'd have issues.

ajkiessl:
After a little more discussion with Nicole, I am going to commit to your solution of restricting the export to only one, and storing the `exported_to_activity_insight` boolean to do this.  Pubs exported from RMD to AI should be published already, so no need to update them.  I would also like to avoid the primary key dependency.

EricDurante:
@ajkiessl We could assign UUIDs to new publications in RMD so that we didn't have to rely on the `id` primary key.

Something like:

```ruby
Class Publication < ApplicationRecord
  before_validation :set_uuid

  private

  def set_uuid
    self.uuid ||= SecureRandom.uuid
  end
end
```

Then we'd have to touch each existing publication record to generate a UUID for each.

I don't think that'll help with _preventing_ duplication in AI, but it should make any future updating of records more reliable, and it would help with _identifying_ duplicates in AI if they were created by RMD.

ajkiessl:
@EricDurante Ok, so that just reminded me: we need some way to match up the record exported to AI from RMD _back_ to the original record when we re-import from AI to RMD.  Otherwise we will end up with a lot of duplicates from the AI import.  So a UUID could work, but then I suppose we could just use the RMD database ID and avoid adding the logic for a UUID.  We would still only allow the export to AI once.  So this is kind of a combination of the two solutions: we only allow export to AI once, we add a field in AI that holds the RMD ID (not as the primary key, though), then when importing from AI use the RMD IDs to update the original record in RMD.  Do you think using a UUID would be better that using the RMD database ID?

EricDurante:
@ajkiessl Yeah, that all sounds like it could work. In RMD we'd definitely want a way to identify incoming data from AI that had been previously exported from RMD.

If we don't have a specific use case for UUIDs, then I'd say we don't need to use them. I can't really imagine a scenario where using the RMD database ID to identify records would be problematic.

ajkiessl:
Waiting for `RMD_ID` field to be added to Activity Insight.

ajkiessl:
RMD_ID is added to AI.  Will start testing this when I finish up my other PR #137

ajkiessl:
It might be better to totally skip any record that has an RMD ID on import. Since the original record will likely be coming from Pure, we don't want cyclical imports of a record from Pure, then shifting back to AI, then back, etc. We can still create an PublicationImport record, but only link it to the underlying publication record. Do not update the actual record.

PR #157, opened 2021-04-05
Stat snapshot fix

ajkiessl:
Limit statistics snapshot to articles only.


ajkiessl:
This is ready to go.

EricDurante (approved):
This looks great, @ajkiessl 
Thanks for doing this!

EricDurante:
I went ahead and deployed this so that tomorrow's snapshot will be correct.

EricDurante:
It looks like the import of other publication types was deployed on 2021-03-11, so I also went ahead and deleted the 4 snapshots that had been saved since then on March 12, March 19, March 26, and April 2.

Issue #120, opened 2021-01-11
Automatically remove known non-duplicates with manual override


ajkiessl:
If an admin is trying to group duplicates that have been marked as non-duplicates, have a dialog pop up that indicates these are non-duplicates and by continuing you will be overriding that.

EricDurante:
@ajkiessl Just a heads-up if you were planning to start working on this issue at some point:

I might be making some significant changes to some of the same code and tests that you'd be touching in order to implement this. To avoid conflicts that might be difficult to resolve, we might want to do this one in series with what I'm currently working on if it can wait.

ajkiessl:
@EricDurante Yeah no problem.  I hadn't planned on working on this just yet.

ajkiessl:
Done in #158

PR #158, opened 2021-04-05
Override known dups

ajkiessl:
Detects non-duplicate groups when merging duplicate groups and prompts the admin if they are sure and want to proceed.

#120

### DIFF
--- a/app/assets/javascripts/rails_admin/custom/initialize_confirm_modal.js
+++ b/app/assets/javascripts/rails_admin/custom/initialize_confirm_modal.js
@@ -1,0 +1,3 @@
+$(document).ready(function(){
+    $("#confirmModal").modal('show');
+});
--- a/app/controllers/custom_admin/publication_merges_controller.rb
+++ b/app/controllers/custom_admin/publication_merges_controller.rb
@@ -11,14 +11,28 @@ class CustomAdmin::PublicationMergesController < RailsAdmin::ApplicationControll
         return
       end
 
+      if params[:known_non_duplicate]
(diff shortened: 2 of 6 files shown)


ajkiessl:
Still testing somethings.  Not ready to merge.

ajkiessl:
Ok, this is ready to go.

EricDurante (approved):
The only thing that I would say about this that might not be obvious is that I think that it's possible for there to be a situation where you could have a duplicate group which contains some publications that belong to the same non-duplicate group, and that the non-duplicate group could have other publications in it that are not in the duplicate group. So overriding and totally deleting the non-duplicate group (as opposed to just removing the offending publications from it and then only deleting it if it's empty) in that case would remove the indicator that those other publications were determined to not be duplicates of each other. 

I can't say for sure whether or not this actually happens in practice, I just that think it's technically possible. This whole business is so mind-boggling that it's difficult to reason about it.. However, if and when such a non-duplicate group ever got deleted, probably the worst outcome would be that some similar-looking publications that aren't actually duplicates would be grouped as suspected duplicates for a second time, and an admin would have to ignore them for a second time. Probably not a huge deal.

So I haven't tried this out, but apart from that caveat above, I think the code looks good. Let's just keep that caveat in the backs of our minds so that we know what's going on if it does turn out to be an issue, and then we can address it. So feel free to merge this if you're happy with it.

ajkiessl:
Yeah that could happen.  I think the admins would pick up on this quickly if it became common.  I'll keep this in mind in case it does.

PR #159, opened 2021-04-13
Import 'In Press' publications from Activity Insight

ajkiessl:
- Logic in ai importer changed
- Added fixture that is 'In Press'
- Specs added/modified to account for this


ajkiessl:
@EricDurante As far as I can tell, the only logic that needs to be changed for this is in the Activity Insight importer.  We can just treat 'In Press' publications like we treat 'Published' publications in the OA workflow.  I ran the integration a few times locally.  It only adds about 2000 publications.  This is ready to go whenever you are ready.

EricDurante:
Hey @ajkiessl, 
Sorry, I'm trying to understand the bigger picture here, and I posted some questions in this Basecamp thread:  https://3.basecamp.com/4206247/buckets/15288919/messages/3651712147

ajkiessl:
@EricDurante No problem.  I'll see how the thread develops.  Looks like I might need to make some changes.

ajkiessl:
This now contains the logic to restrict Pure publication imports to publications with a status of "Published" or "Accepted/In press".  Currently testing imports locally.

ajkiessl:
@EricDurante This is ready to go.  I ran the imports locally and they work.  This doesn't need to be released to beta if you are busy with beta.

EricDurante (approved):
Looks good to me - thanks, @ajkiessl!

Go ahead an deploy this to production, but skip beta. I do still have a different branch deployed there.

PR #228, opened 2021-10-05
Replace Poltergeist and update all our gems

### LINKED ISSUE #222: Replace poltergeist with webdrivers
With headless options for Chrome and Firefox, it's now possible to test integrated javascript features using these browsers instead of simulating one.

## WDDLL

* remove poltergeist gem
* add selenium-webdriver
* configure Capybara
* verify js-based tests use the new webdriver and non-js tests use the rack-based app
* update developer documentation to remove the phantomjs install requirement

## Notes

The current test suite is using `js: true` to trigger the poltergeist-backed browser. Configuring Capybara to use one of its webdrivers should do the trick. See https://github.com/psu-libraries/scholarsphere/blob/develop/spec/support/capybara_drivers.rb for reference.

awead:
Updating our gems was causing test failures because the existing PhantomJS testing browser was creating a window that was too small. Login links were hidden from direct view in a Bootstrap "pancake" that needed to be clicked in order to become visible again. Since we needed to replace PhantomJS anyway, this removes the poltergeist gem and replaces it with Selenium's webdrivers which use the latest Chrome browser for testing.

This fixed the initial login links problem, but created a second problem because the new webdrivers gem does not support request headers. To get around that, we create our own webdriver with a custom download directory that we can monitor. When files have finished downloading, we can verify them instead of just checking for the headers.

Fixes #222


awead:
I think this might just be a simple UI issue:
Screen Shot 2021-10-05 at 12 35 22 PM

ajkiessl (approved):
This looks good to me.

awead:
@ajkiessl I made some changes so it would deploy to qa. That's now working.

EricDurante (approved):
Thank you for all of the modernization/improvement efforts @awead! 👏

awead:
You're welcome, @EricDurante !

PR #229, opened 2021-10-06
OAB publication importer: update publications with existing OA URLs

### LINKED ISSUE #190: OA URL cannot be updated on Admin
The admin UI does not have the ability to update the open access url that is obtained from Open Access Button. In some cases these values are wrong and it would be helpful if the admins could update (or delete) the URL.

nicktiberi:
Closes #190.

The logic for `import_all` has been modified to now query for all publications that have a DOI, regardless of whether they have an open access URL or not. This will have the effect of allowing the OAB publication importer to correct any incorrect URLs that we may have previously received from the OAB API.

### DIFF
--- a/app/importers/open_access_button_publication_importer.rb
+++ b/app/importers/open_access_button_publication_importer.rb
@@ -24,7 +24,7 @@ class OpenAccessButtonPublicationImporter
   private
 
   def all_pubs
-    Publication.where.not(doi: nil).where.not(doi: '').where(%{open_access_url IS NULL OR open_access_url = ''})
+    Publication.where.not(doi: nil).where.not(doi: '')
   end
 
   def new_pubs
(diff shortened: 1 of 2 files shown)


app/importers/open_access_button_publication_importer.rb
```diff
@@ -24,7 +24,7 @@ def import_new
   private
 
   def all_pubs
-    Publication.where.not(doi: nil).where.not(doi: '').where(%{open_access_url IS NULL OR open_access_url = ''})
+    Publication.where.not(doi: nil).where.not(doi: '')
```
awead, line 27: @nicktiberi Do the tests break if you put the old sql statement back in? I'm seeing that you changed the tests, but I don't see the underlying test data has changed. I'm wondering if we could get a regression bug here.

awead (changes requested):
Just a question... otherwise, LGTM.

app/importers/open_access_button_publication_importer.rb
nicktiberi: Yes, they do. The tests have some code that populates the publication with an open access URL:

```
before { pub.update_attribute(:open_access_url, "existing_url") }
```

With the deleted WHERE clause put back, these tests would fail because `all_pubs` would not include the test publication, since it would have a populated `open_access_url`.

EricDurante (approved):
looks good

PR #232, opened 2021-10-07
Rename Capistrano deployments

awead:
Bring our Capistrano environment names inline with our servers. QA now corresponds to the qa server, and staging likewise. The Rails environments are unchanged, so there's still a little mismatch there, but the cap commands should match up with our current communication of deployment environments.


awead:
Right, I was going to go for that too but that was too much work 😆

PR #233, opened 2021-10-08
Add niftany

### LINKED ISSUE #202: Add Niftany
We have the gem "Niftany" which we use to wrap up all our linters such as Rubocop, Eslint, and sass-lint. It's a "nice to have" because it makes linting all your code a one-step process.

## WDLL

- [x]  install Niftany
- [x]  run a basic update and use only "auto-fixable" options
- [x]  mark everything else "todo"
- [x]  add a CI step that fails builds which don't pass linting

banukutlu:
Closes #202 

Installs Niftany, ran auto-correct options with rubocop, erblint and scsslint, excluded everything else and couple cops that failed the tests (left commits unsquashed to show which ones I had to exclude, so better to squash and merge), also added Niftany to CI


EricDurante:
I'm not really familiar with Rubocop, but this generally looks fine to me.

I have pretty high confidence in the test suite to let us know if anything is broken. The only place where I feel like test coverage could be significantly better (this is something that's good to for everyone to know in general) is for some of the non-custom functionality in the admin UI. A lot of the admin functionality and user interface is provided by the rails_admin gem, and we have some basic integration test coverage that verifies that the basic functionality of the admin UI is working, but there's little or no coverage for some of the fancier functionality and UI elements provided by the gem (i.e. searching and filtering). Of course, I'm talking about test coverage for code that we didn't write, but it's functionality that could nevertheless be broken by dependency upgrades and changes to our code.

The only thing that I'd bring up for debate is whether we maybe want to exclude existing database migrations.

banukutlu:
I am working on adding simplecov, will put the PR up soon but just checked on local we are at 96.54% coverage

PR #235, opened 2021-10-08
Add code coverage with simplecov

### LINKED ISSUE #203: Add code coverage with SimpleCov and CodeClimate
A lot of our projects use SimpleCov to calculate code coverage for our tests, along with CodeClimate (CC), which reads these reports and can block PRs that don't match a certain criteria.

CC is optional, if we don't want to get too crazy, but having SimpleCov there is nice so you can see where your tests might be missing things.

## WDLL

* install simple-cov gem
* ensure coverage reports get written after rspec tests
* integrate with code-climate (optional)

banukutlu:
Closes #203

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,26 +54,9 @@ jobs:
       - run:
           name: Niftany
           command: bundle exec niftany
-      # Run the unit tests
       - run:
-          name: Unit Tests
-          command: bin/rspec spec/unit
-
-      # Run the component tests
-      - run:
-          name: Component Tests
-          command: bin/rspec spec/component
-
-      # Run the request tests
-      - run:
-          name: Request Tests
-          command: bin/rspec spec/requests
-
-      # Run the integration tests
-      - run:
-          name: Integration Tests
-          command: bin/rspec spec/integration
-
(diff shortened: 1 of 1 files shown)


awead:
Works for me. We could update the config:

``` ruby
SimpleCov.start 'rails' do
  add_filter('lib/mailer_previews')
end
```

I don't think care about the mailer previews.

EricDurante:
Hmm, something certainly seems to be up with Code Climate's analysis of test coverage..

Also, just a note about the file lib/utilities/pure_downloader.rb that has no test coverage:  that file is sort of dead code. Its functionality has been folded into the Pure importer (which is tested). The only reason I left it in the repo is that it occasionally comes in handy in development when you want to grab some raw data from Pure's API to inspect locally. Absolutely nothing depends on it though.

ajkiessl:
Yeah something's not right with code climate's calculation.  I think we need to use their summing command to combine the outputs of the each test.  There's something like this in ETDA Workflow: https://github.com/psu-stewardship/etda_workflow/blob/main/.circleci/config.yml#L131

awead (approved):

PR #236, opened 2021-10-08
Add Penn State identity information to users

awead:
Integrates the Penn State identity management API into the User record. The response from the API is stored as json in a column of the user table. When retrieved, a PsuIdentity::SearchService::Person is returned, containing additional metadata about the user such as their preferred names, email, and affiliations.

This data can be used to determine if the user is still active at Penn State.

## Details

If the user is present in the Penn State system, updating the data from the API will return a Peson record from the PsuIdentity gem:

``` ruby
>  user = User.find(1)
=> User
>  user.update_psu_identity_data
=> true
>  user.psu_identity.class
=> PsuIdentity::SearchService::Person
>  user.psu_identity.affiliation
=> ["STAFF"]
```

We can use the data as we need without having to re-query the API each time. A nil value would indicate the user doesn't have any records with Penn State.

## Issues

Note that the underlying data in the attribute/column is in json, but we override this to return an object. I don't know if this is a bad convention or not, so I'm putting this out here now for discussion. Additional PRs will build off of this if we like the pattern.


EricDurante:
I don't see any problem with it. Either way I imagine we'll hide the internals from `User`'s public interface.

EricDurante (approved):

awead:
The only thing that crossed my mind was that `psu_identity` is an attribute in the model, so something weird _might_ happen via forms or other magical Rails processes, but it's probably just a matter of ensuring we don't change the value of the attribute any other way except running `update_psu_identity`. Anyway, we can talk about it more at standup next week.

rschenk:
I think this is fine. Personally I don't see this as any "weirder" than a method that returns an ActiveRecord object. As Eric mentioned, if we can disable the writability of this column from Rails Admin, that might be a nice protection. 

I suppose if we were really worried about it, we could name the method that creates the `PsuIdentity::SearchService::Person` object something different than the column that backs it, but I don't feel very strongly about that. I say :shipit:

rschenk (approved):

PR #237, opened 2021-10-08
192 unpaywall data model

EricDurante:
If this looks ok to everyone, we can merge in these schema changes now to maybe save a little pain later.


ajkiessl (approved):
Looks good.  

I think we could potentially split the next steps into a couple different tickets.  Maybe one ticket for building the importer, another for altering the application logic to stop using the old open access urls (this might be pretty involved), and another for the migration script?

PR #243, opened 2021-10-11
Update bundler to version 2

### LINKED ISSUE #204: Update bundler to version 2
Version 1 is outdated. Version 2 has been out for a while and the default for newer installs.

nicktiberi:
Closes #204.

Unless I'm missing something, I think we just needed to run a few upgrade commands to update your local environment:

```
gem install bundler
bundle _2.2.29_ update --bundler
```

This updates the version of `bundler` in `Gemfile.lock`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -597,4 +597,4 @@ DEPENDENCIES
   webdrivers
 
 BUNDLED WITH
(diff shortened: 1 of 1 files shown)


ajkiessl:
Agreed.

PR #248, opened 2021-10-12
Add darwin-19 version of libv8

rschenk:
Ryan's computer (Intel Core i5 MacOS 11.6 20G165) really, really wants to install this version of libv8 every time he bundles


ajkiessl:
I think @awead 's solution is preferable.  Then we don't need to push unecessary packages out to the linux servers.  Assuming a package would never have the word 'darwin' in it outside of indicating the OS.

PR #249, opened 2021-10-12
197 inpress no email

ajkiessl:
Excludes 'In Press' articles from OA workflow and OA emails.  Also restricts the values for publication statuses to 'In Press' and 'Published'.  Maps 'Accepted/In Press" to 'In Press' when updating Publication status.  Any case variation of 'In Press' and 'Published' are mapped to 'In Press' an 'Published' respectively.


app/decorators/authorship_decorator.rb
```diff
@@ -19,20 +19,24 @@ def profile_management_label
...
+          else
+            'question'
+          end
         end
+      else
+        'unlock-alt'
       end
     else
-      'unlock-alt'
+      'circle-o-notch'
```
awead, line 39: You could use some guard clauses to remove the nested if statements:

``` ruby
def open_access_status_icon
  return 'circl-o-notch' unless published?
  return 'unlock-alt' if preferred_open_access_url.present?

  if scholarsphere_upload_pending?
    'hourglass-half'
  elsif scholarsphere_upload_failed?
     'exclamation-circle'
   elsif open_access_waived?
      'lock'
   else
       'question'
   end
end
```

app/mappers/status_mapper.rb
```diff
@@ -0,0 +1,15 @@
...
+    if status.nil?
+      status.to_s
+    elsif status.casecmp('in press').zero? || status.casecmp('accepted/in press').zero?
+      'In Press'
+    elsif status.casecmp('published').zero?
+      'Published'
+    else
+      status.to_s
+    end
+  end
```
awead, line 14: This might be simpler:

``` ruby
    status ||= ''                                                                                                          
                                                                                                                           
    case                                                                                                                   
    when status.match?(/published/i)                                                                                       
      'Published'                                                                                                          
    when status.match?(/in press/i)                                                                                        
      'In Press'                                                                                                           
    else                                                                                                                   
      status.to_s                                                                                                          
    end 
```

You also might not want to call it `map` because that implies an array operation.

app/models/publication.rb
```diff
@@ -19,6 +19,10 @@ def self.publication_types
     ]
   end
 
+  def self.statuses
+    ['Published', 'In Press']
```
awead, line 23: You could define these as constants:

``` ruby
PUBLISHED_STATUS = 'Published'
```

And the reference them elsewhere, like in StatusMapper, as `Publication::PUBLISHED_STATUS`

app/models/publication.rb
```diff
@@ -67,14 +73,16 @@ def self.publication_types
...
-  scope :subject_to_open_access_policy, -> { journal_article.where('published_on >= ?', Publication::OPEN_ACCESS_POLICY_START) }
+  scope :subject_to_open_access_policy, -> { journal_article.published.where('published_on >= ?', Publication::OPEN_ACCESS_POLICY_START) }
 
   scope :open_access, -> { where(%{(open_access_url IS NOT NULL AND open_access_url != '') OR (user_submitted_open_access_url IS NOT NULL AND user_submitted_open_access_url != '') OR (scholarsphere_open_access_url IS NOT NULL AND scholarsphere_open_access_url != '')}) }
 
   scope :journal_article, -> { where("publications.publication_type ~* 'Journal Article'") }
 
   scope :non_journal_article, -> { where("publications.publication_type !~* 'Journal Article'") }
 
+  scope :published, -> { where("publications.status = 'Published'") }
```
awead, line 84: Same here: `Publication::PUBLISHED_STATUS`

app/models/publication.rb
```diff
@@ -577,6 +589,10 @@ def preferred_publisher_name
     preferred_journal_info_policy.publisher_name
   end
 
+  def published?
+    status == 'Published'
```
awead, line 593: ditto here.

app/models/user.rb
```diff
@@ -104,6 +104,7 @@ def self.find_by_nsf_grant(grant)
 
   def self.needs_open_access_notification
     joins(:authorships, :publications, :user_organization_memberships)
+      .where("publications.status = 'Published'")
```
awead, line 107: and here.

awead (changes requested):
Overall this is fine. I suggested a few changes to methods to make them a little less "branch-y". Also, we ought to constantize the various publication statuses instead of passing around their string values. It makes future changes more easy to apply.

app/mappers/status_mapper.rb
awead: Another way to do this:

``` ruby
def self.map(status)                                                                                                     
    case (status || '').to_s                                                                                               
    when /published/i                                                                                                      
      'Published'                                                                                                          
    when /in press/i                                                                                                       
      'In Press'                                                                                                           
    else                                                                                                                   
      status.to_s                                                                                                          
    end                                                                                                                    
  end      
```

PR #254, opened 2021-10-12
Update rails_admin to 2.2.1

nicktiberi:
Updating rails_admin to 2.x.x is a prereq for upgrading to rails 6.


Gemfile
```diff
@@ -43,7 +43,7 @@ gem 'activerecord-import'
 gem 'progressbar'
 
 # RailsAdmin is a Rails engine that provides an easy-to-use interface for managing your data
-gem 'rails_admin', '~> 1.4'
+gem 'rails_admin', '~> 2.2.1'
```
awead, line 46: We could probably be a little more lenient and do 2.2 here, so we could updated to any 2.x version in the future.

awead (changes requested):
One small change, otherwise 👍

awead (approved):

PR #258, opened 2021-10-12
Don't send emails if a user's profile is inactive

awead:
We won't send emails to users if their profile is inactive. We can determine this by looking at the data coming from the identity management API, which is stored in the user record. If there's data there and their affiliation isn't just "MEMEBER" we can be reasonably sure they're still active at Penn State and can receive emails. If that isn't the case, we just don't send the email.

## Details

The UserProfile is going to handle the dirty work. If there's no API data there in the user record, it will update it as soon as the profile is instantiated. We might want to consider updating all the existing user records prior to sending out a mass emails, otherwise we might overload the API. After that point, the hits to the API should be pretty minimal unless we create 100's of new user records in between different mailings.


EricDurante:
This looks good to me. We still have larger, unsolved problems around the idea of a person's "status" at Penn State, so I think that the fact that this acts only as a filter to prevent emails from being sent at the very last step is appropriate. That is to say, I wouldn't necessarily have wanted this data and logic to get mixed up into the definition of who is eligible to receive an open access email or which publications require action for OA policy compliance, etc.

Also, @awead, in case you aren't aware of any or all of this:  in addition to the [html view](https://github.com/psu-stewardship/researcher-metadata/blob/main/app/views/profiles/show.html.erb) of a person's profile that is rendered by the app, there's also an [API endpoint](https://github.com/psu-stewardship/researcher-metadata/blob/26d9f6471797e78d97fca92d3d4bdb641fc21dd2/app/controllers/api/v1/users_controller.rb#L99) for that profile data. That API endpoint, in turn, is used by this whole separate [codebase](https://github.com/psu-stewardship/researcher-metadata-profile-demo) that we have to demonstrate its usefulness. And then an [archive](https://github.com/psu-stewardship/researcher-metadata/blob/main/public/researcher-metadata-profile-demo.zip) of that codebase is also included for download from the Rails app.

awead:
@ajkiessl This is ready to go. I made the change to the API.  I will add the rake task for updating the user's psu_identity field in a separate PR.

ajkiessl (approved):

PR #272, opened 2021-10-14
Log all import errors

### LINKED ISSUE #265: Capture and log errors in all import jobs
In #187 and #188 we developed a pattern to log errors from import jobs, which we implemented on Activity Insight and OAB. This ticket does the same thing for all the importers that are run by cron: 

- [ ] import:rss_news
- [ ] import:ldap
- [x] import:activity_insight  #188 
- [ ] import:pure
- [x] import:open_access_button #187
- [ ] import:scholarsphere

rschenk:
I tried to intelligently handle every error scenario that would be encountered in any of the cron'd import jobs. 

Feel free to squash all these into one. 

Closes #265

### DIFF
--- a/app/importers/pure_journals_importer.rb
+++ b/app/importers/pure_journals_importer.rb
@@ -14,10 +14,19 @@ class PureJournalsImporter < PureImporter
         j.title = item['titles'].first['value']
         j.publisher = Publisher.find_by(pure_uuid: item['publisher']['uuid']) if item['publisher']
         j.save!
+      rescue StandardError => e
+        log_error(e, {
+                    journal_id: j&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+    rescue StandardError => e
+      log_error(e, {})
     end
     pbar.finish unless Rails.env.test?
+  rescue StandardError => e
+    log_error(e, {})
   end
 
   def page_size
--- a/app/importers/pure_organizations_importer.rb
+++ b/app/importers/pure_organizations_importer.rb
@@ -15,8 +15,20 @@ class PureOrganizationsImporter < PureImporter
         o.pure_external_identifier = item['externalId']
         o.organization_type = extract_organization_type(item)
         o.save!
+      rescue StandardError => e
+        log_error(e, {
+                    record_type: record_type,
+                    organization_id: o&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+
+    rescue StandardError => e
+      log_error(e, {
+                  record_type: record_type,
+                  organizations: organizations
+                })
     end
     pbar.finish unless Rails.env.test?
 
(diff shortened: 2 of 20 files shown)


app/importers/pure_journals_importer.rb
```diff
@@ -14,10 +14,19 @@ def call
...
         j.save!
+      rescue StandardError => e
+        log_error(e, {
+                    journal_id: j&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+    rescue StandardError => e
+      log_error(e, {})
```
awead, line 25: Tests aren't hitting this line. I don't know enough to say whether that's important or not.

app/importers/pure_organizations_importer.rb
```diff
@@ -15,8 +15,18 @@ def call
...
+      rescue StandardError => e
+        log_error(e, {
+                    organization_id: o&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+
+    rescue StandardError => e
+      log_error(e, {
```
awead, line 27: Testing miss...

app/importers/pure_organizations_importer.rb
```diff
@@ -33,10 +43,23 @@ def call
...
+        log_error(e, {
+                    child_org_id: child_org&.id,
+                    parent_org_id: parent_org&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+
+    rescue StandardError => e
+      log_error(e, {
```
awead, line 56: Testing miss...

app/importers/pure_publication_tag_importer.rb
```diff
@@ -48,10 +48,21 @@ def call
...
+        log_error(e, {
+                    publication_import_id: pub_import&.id,
+                    publication_id: pub&.id,
+                    fingerprint: fingerprint_xml,
+                    publication: publication
+                  })
       end
       pbar.increment unless Rails.env.test?
+    rescue StandardError => e
+      log_error(e, {})
```
awead, line 61: Testing miss.

app/importers/pure_publishers_importer.rb
```diff
@@ -13,10 +13,20 @@ def call
...
+      rescue StandardError => e
+        log_error(e, {
+                    publisher_id: p&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+
+    rescue StandardError => e
+      log_error(e, {})
```
awead, line 25: Testing miss.

app/importers/pure_user_importer.rb
```diff
@@ -55,10 +55,20 @@ def call
...
+      rescue StandardError => e
+        log_error(e, {
+                    user_id: u&.id,
+                    item: item
+                  })
       end
       pbar.increment unless Rails.env.test?
+
+    rescue StandardError => e
+      log_error(e, {})
```
awead, line 67: Testing miss.

awead (dismissed):
I went through and noted where the tests were missing specific scenarios. I don't know if those particular ones are important or not. If they aren't then go ahead and just "resolve" those and I'll approve this.

I don't know if we're missing any other importer log errors, so maybe another set of eyes will help with that.

app/importers/pure_journals_importer.rb
rschenk: There are a few of these testing misses, which is intentional because due to the way the code is structured it's nearly impossible to have an rspec mock raise an exception that will hit this. Given that this code is simply calling a thoroughly tested method, and since it would be really tricky to pull this off, I didn't think it was worth ending up with a very brittle test suite just to test a method called with no params

banukutlu (approved):
lgtm 👍🏻

PR #285, opened 2021-10-18
Switch from swagger_ui_engine to rswag-ui for API docs

nicktiberi:
`swagger_ui_engine` seems to be abandoned and does not support Rails 6. `rswag` seems to be better maintained and supports Rails 6.

Due to the way that `rswag` generates the HTML for the API docs, I was unable to access many of the helpers that are accessible in "normal" Rails views, such as `link_to`, `image_tag`, and `current_user`. Because of that, I basically just created a new public view, `api_docs.html.erb` and copied a bunch of the generated template over there, so we could reuse the existing app's header (with dynamically appearing Admin link, etc.) and footer.

At this point, all `rswag` is basically doing for us is providing us the `swagger-ui` CSS and JS that we're using in the `api_docs.html.erb` view. We could also probably accomplish this by just using the `swagger-ui-dist` NPM package directly: https://www.npmjs.com/package/swagger-ui-dist

Any preferences? IMO using the NPM package instead of the gem might be a little leaner and preferred, but I am curious what others think.


nicktiberi:
Closing this since we're going the NPM route and using `swagger-ui` directly.

PR #297, opened 2021-10-20
OpenAccessUrlsMigrationService for migrating open access urls from old data model to new data model.

### LINKED ISSUE #259: Migrate open access metadata from old data model to new data model
The `Publication` model has fields called `open_access_url`, `scholarsphere_open_access_url`, and `user_submitted_open_access_url`. We need to copy any data that is present in these fields to the new [open access data model](https://github.com/psu-stewardship/researcher-metadata/issues/192). Whatever mechanism does this copying should look for an existing `OpenAccessLocation` record associated with a given publication which has a `source` that corresponds to the field that is being copied. The URL fields will correspond to the values of the `OpenAccessLocation#source` attribute as follows:
- `open_access_url` -> `'Open Access Button'` (Note:  this won't be correct 100% of the time, but the vast majority of these URLs came from OAB, and we don't have a good way to distinguish those that didn't.)
- `scholarsphere_open_access_url` -> `'ScholarSphere'`
- `user_submitted_open_access_url` -> `'User'`

The process should be idempotent, so if a corresponding `OpenAccessLocation` record already exists, then we should do nothing and move on, otherwise we should create a new record with the appropriate `source` value.

We should consider how we actually want this to be run. For example, it could be implemented as a simple script/rake task, or it could be implemented as an Active Record migration if we want it to be run everywhere automatically (but only once). Depending on how/when we roll the various parts of this Unpaywall epic out to production, it may be useful to be able to run this process more than once.

ajkiessl:
closes #259 

**Do not merge** until Ryan is done with his refactoring, so I can implement his `source` methods.

### DIFF
--- a/app/services/open_access_urls_migration_service.rb
+++ b/app/services/open_access_urls_migration_service.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class OpenAccessUrlsMigrationService
+  def self.call
+    Publication.all.each do |pub|
+      if pub.open_access_url
+        OpenAccessLocation.find_or_create_by(source: 'Open Access Button',
+                                             url: pub.open_access_url,
+                                             publication_id: pub.id)
+      elsif pub.scholarsphere_open_access_url
+        OpenAccessLocation.find_or_create_by(source: 'ScholarSphere',
+                                             url: pub.scholarsphere_open_access_url,
+                                             publication_id: pub.id)
+      elsif pub.user_submitted_open_access_url
+        OpenAccessLocation.find_or_create_by(source: 'User',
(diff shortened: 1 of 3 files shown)


nicktiberi (approved):

app/services/open_access_urls_migration_service.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class OpenAccessUrlsMigrationService
+  def self.call
+    Publication.all.each do |pub|
```
awead, line 5: The `:all` method is like an anti-pattern, it seemed like a good idea at the time but it doesn't perform well. The method loads up all the records into memory. Use `:find_each` instead.

PR #298, opened 2021-10-20
Implement Unpaywall publication importer

### LINKED ISSUE #257: Import open access metadata from Unpaywall
Using the updated data model (https://github.com/psu-stewardship/researcher-metadata/issues/192) we need to be able to automatically query Unpaywall's [API](https://unpaywall.org/products/api) for each of our publications that has a DOI, and store any open access [metadata](https://unpaywall.org/data-format) that is returned. If any metadata exists, it should create/update an `OpenAccessLocation` record that is associated with the publication, and it should also update the `Publication`'s `open_access_status` field.

This import needs to run in such a way that it does not exceed Unpaywall's rate limit of 100,000 requests/day. Like the Open Access Button import, this import should be scheduled to run automatically once per week, and it should continue and log issues after encountering any unexpected errors.

nicktiberi:
**NOTE: ** Don't merge this until Ryan is done with his refactor of Open Access Location strings into enums. This PR uses the 'Unpaywall' string which will need to be changed to fit the refactor.

Closes #257.

This new importer imports Unpaywall's "best" OA URL for publications, as well as the current Open Access status of the publication.


app/importers/unpaywall_publication_importer.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+class UnpaywallPublicationImporter
+  def import_all
+    unless Rails.env.test?
+      pbar = ProgressBar.create(title: 'Importing publication data from Unpaywall',
```
awead, line 6: I've found that running ProgressBar in test is fine, so go ahead and nix the `Rails.env.test?` checks.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def new_pubs
+      all_pubs.where(unpaywall_last_checked_at: nil)
+    end
+
+    def get_pub(url)
+      attempts = 0
+      HTTParty.get(url).to_s
+    rescue Net::ReadTimeout, Net::OpenTimeout
```
awead, line 43: This block isn't getting tested, and seems worthy enough. What do you think? I can give you a hand if you need.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+
+      publication.open_access_status = oa_status
+      publication.unpaywall_last_checked_at = Time.current
+      publication.save!
+
+      # Unpaywall asks that users limit requests to no more than 100,000 per day.
+      # Limiting to 1 request per second caps us at 86,400 requests per day.
+      unless Rails.env.test?
+        sleep 1
```
awead, line 78: Good one-liner here: `sleep 1 unless Rails.env.test?`. 🥁

spec/component/importers/unpaywall_publication_importer_spec.rb
```diff
@@ -0,0 +1,378 @@
+# frozen_string_literal: true
+
+require 'component/component_spec_helper'
+
+describe UnpaywallPublicationImporter, :vcr do
+  let(:importer) { described_class.new }
+
+  let(:now) { Time.new(2019, 11, 13, 0, 0, 0) }
```
awead, line 8: If you need dates in the past, try `Time.zone.now - 2.years`.

spec/component/importers/unpaywall_publication_importer_spec.rb
```diff
@@ -0,0 +1,378 @@
...
+    end
+
+    context 'when an existing publication has a DOI that corresponds to an available article listed with Unpaywall' do
+      let!(:pub) { create :publication,
+                          doi: 'https://doi.org/10.1001/jamadermatol.2015.3091',
+                          unpaywall_last_checked_at: last_check,
+                          open_access_status: 'green' }
+
+      context 'when the publication has been checked in Unpaywall before' do
+        let(:last_check) { Time.new(2021, 1, 1, 0, 0, 0) }
```
awead, line 239: `Time.zone.yesterday` better? This also might make a good factory bot trait.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -0,0 +1,91 @@
...
+          existing_oa_location.update!(url: best_oa_location)
+        else
+          publication.open_access_locations.create!(source: 'Unpaywall', url: best_oa_location)
+        end
+      else
+        existing_oa_location.try(:destroy)
+      end
+
+      publication.open_access_status = oa_status
+      publication.unpaywall_last_checked_at = Time.current
```
awead, line 72: `Time.zone.now`

spec/component/importers/unpaywall_publication_importer_spec.rb
```diff
@@ -0,0 +1,378 @@
...
+        let(:last_check) { nil }
+
+        context 'when the publication has no open access locations' do
+          it 'does not create any new open access locations' do
+            expect { importer.import_new }.not_to change(OpenAccessLocation, :count)
+          end
+
+          it 'updates Unpaywall check timestamp on the publication' do
+            importer.import_new
+            expect(pub.reload.unpaywall_last_checked_at).to eq now
```
awead, line 367: Here you can use something like `be_within(1.minute).of(Time.zone.now)`

app/importers/unpaywall_publication_importer.rb
nicktiberi: 👍 I still sometimes forget about this Ruby way of writing conditionals. 😁

app/importers/unpaywall_publication_importer.rb
nicktiberi: I'm fine with making this switch, but just curious what the benefit is. It looks like we are using both pretty equally throughout the app.

rschenk: @nicktiberi You should _always_ use `Time.zone` in Rails projects, eg `Time.zone.now` and `Time.zone.parse` and `Time.zone.local`. Using `Time.new` will only cause you pain and suffering when all your stuff breaks +/- 5 hours from midnight. 

The reason is that by default Ruby does not take timezones into account, and therefore `Time.new` is blissfully unaware of what a timezone is. Our production servers are running in UTC time, and everything stored in the database is stored in UTC time (as it should be). However, our application code is configured to run in Eastern Time (as it should be). Therefore we need to use `Time.zone` when we work with dates and times, to attach the current timezone to the Datetime object, so it will be properly translated from eastern time, into utc, and back again.

This doesn't cause any issues when you run your tests on your development machine, because your system time zone (Eastern) matches the application's configured time, but in ci and on production this can lead to bugs and broken tests.

tldr; never never never use `Time.new` and `Time.now`, always use `Time.zone._____` in Rails apps

Sincerely, 
Lord of The Time Zones, who once literally got into a fight with NOAA/NASA (and won) about the correct handling of timezones.

app/importers/unpaywall_publication_importer.rb
nicktiberi: Great call to test this, because it was broken 😁 

`attempts` kept getting reset to 0 at the beginning of the function so it would keep attempting the API call in an infinite loop. We'll have to make this same change in the `open_access_button_importer` as well. I will open a ticket for fixing that + other improvements based on the feedback from this PR.

nicktiberi: Yeah, that's a great idea. What is the convention for declaring/importing "utility methods" in Rails? I would think we could rename this method to something like `HTTPHelper::try_get(url)` or something like that. But definitely open to other naming suggestions.

awead (approved):
Nice!

Issue #295, opened 2021-10-19
Refactor PublicationImport and relevant -Importer classes to DRY up magic strings in their `sources`


rschenk:
The problem is discussed at length in #294 and I won't repeat that discussion here.

This ticket will implement one half of the fix, using (hopefully) the patterns established in #294 to fix up `PublicationImport` and any Importer that inherits from `OAIImporter` to use the same implementation, thus avoiding duplication of the strings.

PR #299, opened 2021-10-20
WIP: OpenAccessLocation uses a value object for sources

### LINKED ISSUE #294: Refactor OpenAccessLocation's `sources` to be a value object
Currently, we are using magic strings for sources in several places, take for example:

```
$ grep 'Penn State Law eLibrary Repo' app/**/*.rb
app/importers/psu_law_school_publication_importer.rb:      'Penn State Law eLibrary Repo'
app/models/open_access_location.rb:      'Penn State Law eLibrary Repo'
app/models/publication_import.rb:      'Penn State Law eLibrary Repo',
```

This string (and others like it) is validated against by two different models, `OpenAccessLocation` and `PublicationImport`, which check its exact format and its inclusion in their authorized list of sources. Furthermore it is also used in presentational code which is shown to the user.

This creates a number of sneaky, hidden dependencies that will proliferate more once we start building out #192 
- If one instance of this string needs to change, then all of them need to change, but there are no tests that ensure that. It's a hidden dependency
- These strings both represent a value and the display of that value, which violates MVC. E.g. if PSU Law School gets a big donation and needs to rename itself from "Penn State Law eLibrary Repo" to "Jerome K. Jerome Foundation Virtual E-Reading Room", we have no way to do that without changing several classes and/or a lot of data in the db
- When we start coding up #260 these magic strings will appear in more places in the codebase and now have semantics attached to them as we write logic to rank them by preference

Ultimately what we need to do is refactor these sources into a single "value object" pattern so the magic string is only defined in one place, and then we need to add the ability to change its user facing display too. Ultimately this will be shared between the `PublicationImport`, `OpenAccessLocation` and any `-Importer` classes that tie the two together.

To keep ourselves from getting too much into the weeds, this ticket will cover only the `OpenAccessLocation` side of things, so we can build out the rest of #192 without creating extra work for ourselves. Then we'll take care of the `PublicationImport` and `-Importer` side of the equation.

## Todo

1. #295
2. ???

rschenk:
Changes the use of magic strings used in the `OpenAccessLocation` model, and replaces them with a value object, `Source` which...

- defines the string that is ultimately stored in the database as a constant
- Also defines the human-friendly presentational values, stored en en.yml

The OpenAccessLocation model uses these constants as an `enum`, which
- validates that the attribute is present
- validates that the attribute is in the given list
- creates some scopes for finding/filtering, e.g. `OpenAccessLocation.source_scholarsphere.find_all`
- creates handy predicate methods, e.g. `my_oal_instance.source_scholarsphere?`

By default, Rails enums are stored as integers in the database, which I do not like because it makes your data totally opaque. A small modification allows them to be stored as string values in the db, which I like a lot better. That's what the helper method `to_enum_hash` does.

Closes #294 

The same Source value object and a similar enum can be used over on `PublicationImport` for #295, and you'll see some comments in this PR where I had to tweak a couple private methods to allow this bifurcation. Those will go away once #295 is done.

I see that CI has exploded on the integration specs. I'll update those later, but for now, please take a look at what I have so far

### DIFF
--- a/app/models/open_access_location.rb
+++ b/app/models/open_access_location.rb
@@ -1,30 +1,33 @@
 # frozen_string_literal: true
 
 class OpenAccessLocation < ApplicationRecord
-  def self.sources
-    [
-      'User',
-      'ScholarSphere',
-      'Open Access Button',
-      'Unpaywall',
-      'Dickinson Law IDEAS Repo',
-      'Penn State Law eLibrary Repo'
-    ]
-  end
+  enum source: to_enum_hash([
+                              Source::USER,
+                              Source::SCHOLARSPHERE,
+                              Source::OPEN_ACCESS_BUTTON,
+                              Source::UNPAYWALL,
+                              Source::DICKINSON_IDEAS,
+                              Source::PSU_LAW_ELIBRARY
+                            ]),
+       _prefix: :source
 
   belongs_to :publication, inverse_of: :open_access_locations
 
   validates :publication, :source, :url, presence: true
-  validates :source, inclusion: { in: sources }
+
+  def source
+    @source ||= (Source.new(read_attribute(:source)) if read_attribute(:source).present?)
+  end
 
   def name
-    "#{url} (#{source})"
+    "#{url} (#{source&.display})"
   end
 
   rails_admin do
     show do
       include_all_fields
 
+      field(:source) { pretty_value { Source.new(value).display } }
       field(:oa_date) { label 'OA date' }
       field(:url) do
         label 'URL'
@@ -44,7 +47,7 @@ class OpenAccessLocation < ApplicationRecord
       field(:url) { label 'URL' }
       field(:source, :enum) do
         enum do
-          [value || 'User']
+          [value || Source::USER].index_by { |str| Source.new(str).display }
         end
       end
     end
@@ -53,7 +56,7 @@ class OpenAccessLocation < ApplicationRecord
       field(:url) { label 'URL' }
       field(:source, :enum) do
         enum do
-          [value || 'User']
+          [value || Source::USER].index_by { |str| Source.new(str).display }
         end
       end
     end
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+class Source
+  USER = 'user'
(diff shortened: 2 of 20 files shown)


app/models/open_access_location.rb
```diff
@@ -1,24 +1,26 @@
...
-  end
+  enum source: to_enum_hash([
+                              Source::USER,
+                              Source::SCHOLARSPHERE,
+                              Source::OPEN_ACCESS_BUTTON,
+                              Source::UNPAYWALL,
+                              Source::DICKINSON_IDEAS,
+                              Source::PSU_LAW_ELIBRARY
+                            ]),
+       _prefix: :source
```
awead, line 12: This looks good. What is _prefix and where is to_enum_hash coming from?

awead: I'm assuming it's the same as https://github.com/psu-libraries/scholarsphere/blob/develop/app/models/work.rb#L98

awead: oh, I see... it's in ApplicationRecord.

app/models/open_access_location.rb
```diff
@@ -1,24 +1,26 @@
...
   validates :publication, :source, :url, presence: true
-  validates :source, inclusion: { in: sources }
+
+  def source
+    @source ||= (Source.new(read_attribute(:source)) if read_attribute(:source).present?)
+  end
 
   def name
-    "#{url} (#{source})"
+    "#{url} (#{source&.display})"
```
awead, line 23: This could look like "http:... ()" 🤷

app/models/source.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def ==(other)
+    to_s == other.to_s
+  end
+  alias :eql? :==
+
+  def display
+    I18n.t(
+      @source,
```
awead, line 22: Would this be nil if the attribute wasn't present in OpenAccessLocation?

awead (approved):
This looks good to me. I left a couple of questions.

app/models/source.rb
rschenk: It would be `""` but I could easily make it nil.

Given that these are used in enums, which do not allow it to be nil, this is super unlikely to happen in practice, but yes you could force it to happen

PR #302, opened 2021-10-21
264 swagger UI

### LINKED ISSUE #264: Update swagger_ui_engine
The latest version of `swagger_ui_engine` gem does not support Rails 6. There are a few options for how to proceed:

- replace `swagger_ui_engine` with `rwsag`, which is better maintained and supports Rails 6: https://github.com/rswag/rswag#route-prefix-for-swagger-json-endpoints
- update to the unreleased v1.1.4 of `swagger_ui_engine`: https://github.com/zuzannast/swagger_ui_engine/tree/1.1.4-version
- roll this ourselves without a gem

It seems to me that the `rswag` route might be the best option for long-term maintainability, since `swagger_ui_engine` seems to be abandoned at this point.

nicktiberi:
Closes #264.

Since `swagger_ui_engine` is no longer compatible with Rails 6, we've switched to using the `swagger-ui-dist` NPM package directly.

In order to accomplish this, we implemented webpacker as well as the Stimulus JavaScript library to facilitate loading and using NPM packages within the app.

Screen Shot 2021-10-21 at 11 58 10 AM


app/views/public/home.html.erb
```diff
@@ -31,7 +31,7 @@
       <p>
       We publish clean metadata through our RESTful API adhering to the Open API Standard. Explore our endpoints using our interactive documentation. <a href="mailto:L-FAMS@lists.psu.edu?Subject=RMD%20API%20Key%20Request" target="_top">Request</a> a license key to query the data.
       </p>
-      <%= link_to "API Documentation", swagger_ui_engine_path, class: 'btn btn-primary' %>
+      <%= link_to "API Documentation", '/api_docs', class: 'btn btn-primary' %>
```
awead, line 34: Use `api_docs_path` here

ajkiessl:
Might want to try deploying this to qa before merging to see if we need to add anything to the deploy.rb.  We might need to add a line to run yarn.

Issue #292, opened 2021-10-19
Importers do not call progressbar unless `$stdout.tty?`


rschenk:
Launching an importer into the background will cause an error explosion 

Steps to reproduce bug: 
- Start `bundle exec rake import:psu_identity`
- ctrl+z then `bg` and `disown`
- You can reproduce it locally if you have a 100 or so user records, just add the same person 100 times

PR #325, opened 2021-10-29
Do not render progress bars in test mode, or non-interactive terminals

rschenk:
I took a crack at #292 and wrapped the`ProgressBar.create` factory in a little wrapper that tests if the current stdout is not a tty or if we are running in test mode. If either of those conditions are met, then it will use the progressbar's handy built-in Null Object for its  output instead of stdout.

There is a catch, though. The way that Adam wanted to use this in #292, by launching the script in the foreground and then backgrounding it, will not work, because the `tty?` check is only performed at initialization time. You could however do: `$ nohup bundle exec rake import:psu_identity &> /dev/null &` which would take care of things.

The above makes me think that this is ... less good than I was hoping it would turn out, but I still thought I would put it up for review before sinking any more time into it

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -10,10 +10,10 @@ class ActivityInsightImporter
   attr_reader :errors
 
   def call
-    pbar = ProgressBar.create(title: 'Importing Activity Insight Data', total: ai_users.count) unless Rails.env.test?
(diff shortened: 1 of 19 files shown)


awead (approved):
Nice!

awead:
Using `nohup` is probably more correct because that's what you'd expect someone to do. Trying to detect backgrounding, etc, is overthinking it.

ajkiessl (approved):

PR #327, opened 2021-10-29
endpoint to update SS OA Link

### LINKED ISSUE #196: API to allow update to SS OA Link
The product owner for ScholarSphere should be able to update the ScholarSphere Open Access Link property in RMD to allow for scripting of uploads to ScholarSphere that are in compliance with OA Policy (when the upload takes place outside RMD integration).

banukutlu:
Closes #196


app/controllers/api/v1/publications_controller.rb
```diff
@@ -30,6 +30,18 @@ def grants
       render json: API::V1::GrantSerializer.new(publication.grants)
     end
 
+    def update_all
+      if api_token.write_access?
+        if update_ss_oa_url.positive?
+          render json: { message: I18n.t('api.publications.patch.success'), code: 200 }, status: 200
+        else
+          render json: { message: I18n.t('api.publications.patch.no_content'), code: 404 }, status: 404
```
awead, line 38: You would get this response if one of the updates failed. And, in that case, any other updates would be rolled back. I'm wondering if the message should say something to that effect. I don't know who the intended user is for this endpoint. Would they want to know that the updates didn't work? Maybe it doesn't matter... but I'm just bringing up the point here. If the updates are rolled back, do we care?

awead (changes requested):

app/controllers/api/v1/publications_controller.rb
```diff
@@ -226,6 +238,72 @@ def grants
...
+          end
+        end
+        response 200 do
+          key :description, 'publications update success response'
+          schema do
+            key :'$ref', :PublicationsPatchResult
+          end
+        end
+        response 404 do
+          key :description, 'unauthorized'
```
ajkiessl, line 269: I think this should be 'not_found' or something along those lines.

banukutlu: yes that is right, nice catch! I will fix that.

app/controllers/api/v1/publications_controller.rb
EricDurante: I think it would also be a little unclear to me as an API user what would happen if I provided both an Activity Insight ID and a DOI in either the case where those parameters were for the same publication or the case where they were for different publications. It seems to me like the intention is to update one representation of a publication at a time, and so it would make more sense to me if a user could provide _either_ a DOI _or_ an Activity Insight ID, but not both. Right now it looks like our code will accept both parameters at once for two different publications, and it will update both publications with the same URL. I kind of think we should change that so that it's only one of those parameters per request. And then I think all that we could really do to make this behavior more clear to users is to be explicit about it in the documentation and then respond with an error if both params are sent at once.

EricDurante (changes requested):
@banukutlu Sorry for all of the feedback/comments on this one. I think we're sort of still in the process of iterating on the design of this as you're writing it. Since it's the first one, this will probably set somewhat of a design precedent for any future write operations that happen via this API, so I think it's good for all of us to mull it over, try things, and see what we like and don't like.

app/controllers/api/v1/publications_controller.rb
banukutlu: fixed with the last commit

app/controllers/api/v1/publications_controller.rb
```diff
@@ -226,6 +238,72 @@ def grants
...
+        key :operationId, 'updatePublications'
+        key :produces, [
+          'application/json',
+          'text/html'
+        ]
+        key :tags, [
+          'publication'
+        ]
+        parameter do
+          key :name, :Publications
```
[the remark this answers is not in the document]
banukutlu: updated with the last commit.

app/controllers/api/v1/publications_controller.rb
```diff
@@ -226,6 +236,92 @@ def grants
...
+          end
+        end
+        security do
+          key :api_key, []
+        end
+      end
+    end
+
+    swagger_schema :PublicationInput do
+      one_of do
```
banukutlu, line 290: @EricDurante I use `one_of` here to define the schema thinking that the input would reflect as either containing a doi or activity_insight_id but did not work as the way I expected. Any thoughts on how to define the schema here?

awead: @banukutlu I think I remember running into this situation before when I was writing docs for SchlarSphere's API. I gave up trying to figure how how to make it "one of these" and just put in a descriptive note. I think if the error message is clear, "You provided both A and B, but you can only use A or B" or something along those lines.

banukutlu: I am kinda at that point, I tried couple variations and also oneOf is not supported by 2.0. 

So, should we just leave the doc as below in that case? I am not sure how to represent it 🤔 

```
{
 "doi": "string"
  "activity_insight_id": "string",
  "scholarsphere_open_access_url": "string"
}
```

EricDurante: @banukutlu 
Yeah, I totally agree with Adam. I would maybe just note this restriction in the description of the publication parameter and then also make sure that the error message is descriptive like Adam suggested. I think that will be enough to clarify the proper usage.

app/controllers/api/v1/publications_controller.rb
```diff
@@ -253,5 +337,65 @@ def filter_by_doi(query)
...
         query.where(doi: doi)
       end
+
+      def params_valid?
+        ai_id = params[:activity_insight_id]
+        doi = params[:doi]
+
+        (ai_id.present? && doi.nil?) ||
+          (ai_id.nil? && doi.present?) &&
+            params[:scholarsphere_open_access_url].present?
```
awead, line 347: if both ai_id and doi are present, it's false, so we could rewrite it:

``` ruby
def params_valid?
  return false unless params[:scholarsphere_open_access_url].present?

  !(params[:activity_insight_id].present? && params[:doi].present?)
end
```

awead (changes requested):
Overall, I think this is fine, but it could use some clarification. Another option that might make life easier to make the whole thing its own service or Ruby object. For example, all the updating, parsing, etc. could be in a single class, and then the update_all becomes:

``` ruby
def update_all
  if api_token.write_access?
    operator = UpdateAllPublicationsOperator.call(params: params)
    render json: { message: operator.message, code: operator.code}, status operator.code
  else
    render json: { message: I18n.t('api.errors.not_authorized'), code: 401 }, status: 401
  end
end
```

awead:
Or better yet:

``` ruby
def update_all
  operator = UpdateAllPublicationsOperator.new(params: params, token: api_token)
  operator.perform
  render json: { message: operator.message, code: operator.code}, status operator.code
end
```

app/importers/unpaywall_publication_importer.rb
```diff
@@ -107,6 +107,6 @@ def update_open_access_location(open_access_location, unpaywall_json)
     end
 
     def build_new_oal(publication, url)
-      publication.open_access_locations.build(source: Source::UNPAYWALL, url: url)
+      publication.open_access_locations.build(source: Source::USER, url: url)
```
ajkiessl, line 110: This should stay as `Source::UNPAYWALL`, right?

ajkiessl (changes requested):

app/importers/unpaywall_publication_importer.rb
banukutlu: oh yes! thanks for catching this, totally mistakenly changes.

banukutlu: fixed now.

app/controllers/api/v1/publications_controller.rb
banukutlu: yes 👌🏻 ☝🏻

ajkiessl (approved):
Looks good!

banukutlu:
@ajkiessl @awead I did some refactoring to take some of the logic out of the controller, would you mind reviewing that as well?

app/controllers/api/v1/publications_controller.rb
```diff
@@ -10,11 +10,11 @@ def index
       query = api_token.publications.visible.limit(limit)
 
       if params[:activity_insight_id].present?
-        query = filter_by_activity_insight_id(query)
+        query = Publication::filter_by_activity_insight_id(query, params[:activity_insight_id])
       end
 
       if params[:doi].present?
-        query = filter_by_doi(query)
+        query = Publication::filter_by_doi(query, params[:doi])
```
awead, line 17: `Publication.filter_by_doi`

app/services/scholarsphere_location_operator.rb
```diff
@@ -0,0 +1,86 @@
...
+        source: Source::USER,
+        url: params[:scholarsphere_open_access_url]
+      )
+    end
+
+    def operator_response(code, message)
+      OperatorResponse.new(code: code, message: message)
+    end
+
+    class OperatorResponse
```
awead, line 78: This can just be a struct, but I don't think there's any disadvantage to the way you have it here.

banukutlu: Interesting, I started out with a struct then turned it into a class when I thought I would add couple other methods but eventually I did not need those additinal methods so yeah it can actually be just a Struct. I will change it.

ajkiessl (approved):

app/services/scholarsphere_location_operator.rb
banukutlu: @awead I changed it to an OpenStruct.

awead (approved):

banukutlu:
@EricDurante would you be able to do a quick review on this, I'd made some changes that might be addressing your earlier suggestions. The PR is in requested changes status w/ your review, I can still merge it if you do not have time but checking to see if you can.

app/services/scholarsphere_location_operator.rb
```diff
@@ -0,0 +1,77 @@
...
+    def existing_location?(publication)
+      @existing_location ||= publication.open_access_locations
+        .filter { |l| l.source == Source::USER }
+        .index_by(&:url)
+        .key?(params[:scholarsphere_open_access_url])
+    end
+
+    def find_or_create_location(publication)
+      publication.open_access_locations.find_or_create_by(
+        source: Source::USER,
```
EricDurante, line 69: Here and in the method above I think we should be using `Source::SCHOLARSPHERE` instead of `source::USER`. `Source::USER` is for `OpenAccessLocation`s that are created by faculty/researcher users submitting a URL via the web form. It does feel a little odd to create an `OpenAccessLocation` with `Source::SCHOLARSPHERE` here since it could be any API user submitting the request and not necessarily the ScholarSphere app (or someone acting on its behalf). However, the param that's being submitted is `scholarsphere_open_access_url`, and in the old data model we would have been pushing the value into the field of the same name on the `Publication` which was also used by the ScholarSphere importer and the ScholarSphere upload workflow in RMD. In the new data model, instead of a value in that field, we'd have an associated `OpenAccessLocation` with a `source` of `Source::SCHOLARSPHERE` and a value in its `url` field.

So we'll probably be overloading the meaning of `Source::SCHOLARSPHERE` a little bit, but I'm not sure that it really matters. We could introduce a new `source` value like `Source::API_USER` or something like that, but it would complicate this even further and probably wouldn't be necessary.

EricDurante (changes requested):
@banukutlu I just have one more minor change request to approve this, but your other changes look good to me. Thanks for your patience working through the design of this feature!

app/services/scholarsphere_location_operator.rb
```diff
@@ -0,0 +1,89 @@
...
+      elsif params[:doi].present?
+        publications = Publication.filter_by_doi(Publication, params[:doi])
+      end
+
+      publications
+    end
+
+    def existing_location?(publication)
+      @existing_location ||= publication.open_access_locations
+        .filter { |l| l.source == Source::USER }
```
EricDurante, line 74: @banukutlu 
I'm sorry, I meant that I think that we need that source change in this spot too. `Source::SCHOLARSPHERE` is the value that the ScholarSphere importer and upload workflow are going to be using, so we probably won't find any existing open access locations if we use `Source::USER` here.

banukutlu: yes, you are right! thanks for catching it, I missed it :)

EricDurante (approved):

PR #331, opened 2021-11-02
Update controllers to use UserController

### LINKED ISSUE #322: Admin masquerade updates to profile publication order and visible boolean
Currently when masquerading as another user, changing the order of the publications in the user's profile does not save.  Also, checking and unchecking "Visible in profile" does not save.

These two things need to be saved when changed.

awead:
Controllers that perform `authenticate_user!` need to consistently set the `current_user`, whether this is the actual user, or an admin impersonating them. The inheritances are shuffled a bit so that any controller authenticating a user will subclass `UserController`. There are also additional refactorings, bringing all the unauthenticated controller tests into on shared test.
Controllers that subclass `UserController` just have a simple type check. Then the actual user controller spec test verifies that the builder class is used to set the current user. 

Although this fixes the underlying issue in this ticket, there aren't any integration tests that verify it because the only thing that's changing is the current user. Since the builder class tests that thoroughly, and we now verify that the other controllers will use that class, further integration tests seem unnecessary.

Fixes #322

### DIFF
--- a/app/controllers/authorships_controller.rb
+++ b/app/controllers/authorships_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-class AuthorshipsController < ApplicationController
-  before_action :authenticate_user!
-
+class AuthorshipsController < UserController
   def update
     authorship = current_user.authorships.find(params[:id])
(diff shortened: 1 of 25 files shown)


EricDurante (approved):
This looks great :shipit:

PR #333, opened 2021-11-02
Switch to using our forked version of apivore

### LINKED ISSUE #287: Update apivore gem or find alternative
The `apivore` gem depends on `actionpack` < 6, while Rails 6 depends on `actionpack` 6, so we may need to either fork it or find an alternative. 

`apivore` has not been updated since 2017. There is an unmerged PR which seems to have successfully updated it to support Rails 6. Not sure if we want to just use this patch, or if we would prefer to find an alternative.

We are currently using `apivore` here to test the API: https://github.com/psu-stewardship/researcher-metadata/blob/main/spec/requests/api/v1/swagger_checker_spec.rb

nicktiberi:
Closes #287. 

This resolves our last dependency incompatibility with Rails 6. I figured I would get this merged in its own PR since the Rails 6 upgrade PR looks like it will have a lot of changes.

We've made some minor dependency changes to allow it to be compatible with Rails 6.


nicktiberi:
For some reason, CI can't check out our fork of apivore from GitHub even though the repo is public. Looking into why that might be...

EricDurante:
It's not obvious to me what's going wrong on CI. Have you tried just rerunning it? Maybe it was just some temporary hiccup.

nicktiberi:
@EricDurante - I just added `.git` to the repo URL and that seemed to have fixed it. It was working fine locally without it, but CI couldn't find the repo without it for some reason.

EricDurante:
@nicktiberi Weird. I was actually going to suggest trying that next, but I had absolutely no basis for making the suggestion - it was just a guess. 😆 🤷‍♂️

EricDurante (approved):

PR #334, opened 2021-11-02
Allow any Penn State user to create an account

### LINKED ISSUE #267: Allow students and staff to create their own accounts in RMD
Allow any Penn State user to create an account in RMD.

## WDDLL

* integrate with Azure to effect a login feature exactly like we do with Scholarsphere
* create new User record populated with current metadata
* update psu_identity field with current data from the API
* default to "no organizations"
* verify that these users can use the external waiver form

awead:
After authenticating with Azure, if the user is not present in RMD, an account will be created for them using their current identity data from Penn State. Only the minimal amount of information will be stored, and the user will have no organizational memberships.

I have verified that users can submit open access waiver forms via the Publications tab, but there are no additional integration tests needed for that because we don't distinguish between users created via import or users created via direct login.

FIxes #267

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -46,16 +46,7 @@ class User < ApplicationRecord
   accepts_nested_attributes_for :user_organization_memberships, allow_destroy: true
 
   def self.from_omniauth(auth)
-    # We've added `uid` and `provider` fields to this model to support
-    # multi-provider omniauth, but in reality, we're only using one provider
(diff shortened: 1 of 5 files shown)


awead:
Coverage looks fine, but for some reason analysis isn't working https://codeclimate.com/github/psu-stewardship/researcher-metadata/builds/193

I'm happy to ignore that and move on. Hopefully it will resolve itself soon.

ajkiessl (approved):
Looks good.

PR #337, opened 2021-11-03
Remove remnants of old open access data model

### LINKED ISSUE #261: Clean up remnants of old open access data model
Once the rest of the epic https://github.com/psu-stewardship/researcher-metadata/issues/192 is complete, we can drop the deprecated `Publication` fields, `open_access_url`, `scholarsphere_open_access_url`, and `user_submitted_open_access_url` (and their data) from the database.

rschenk:
The only thing really necessary here was to drop the actual database columns, remove a couple fields from the RailsAdmin config, and then fix a few specs that we accidentally missed in #311 

Closes #261


awead (approved):
It's always a good sign to the see that the only we needed to do was delete the column and update the spec tests.

EricDurante (approved):
Looks good

PR #340, opened 2021-11-04
Tasks to pull data from production and load the data into local, qa, and stage envs

### LINKED ISSUE #336: Task/script for importing prod db data into qa or staging

ajkiessl:
This contains a task to pull the db data file from production, and separate tasks for loading that file to the local db, qa db, and stage db.  There is some duplication in here, but I didn't feel like it was enough to warrant abstracting this into a class or module.  I could try to do that if anyone feels that should be done.

closes #336


awead (commented):
I wonder if using Net::SSH might be better here. Note: I haven't done this before. Looking at the docs: https://www.rubydoc.info/github/net-ssh/net-ssh/Net/SSH/Connection/Session#exec-instance_method

One thing that jumps out at me is that it handles error cases really well.

You could also, in theory, write a Capistrano task to do this since you're basically authenticating using the same user, and the host information is already in the deploy/prod.rb config.

In either case, what you have works, so I won't block this, but I'm just throwing out an alternative if that's easier.

EricDurante:
I think Adam's suggestions are fine, but I don't see any need to change this if it already works well enough in practice. 

Ultimately I'd feel a little better if we had separate scripts for dumping and loading the databases that lived on the servers (and in the chef codebase) so that all that the remote script would need to do is to invoke the server-side scripts and then move files around, but this is good for now.

ajkiessl:
I played around with Net::SSH a little.  I'm going to keep it the way it is for now.

lib/tasks/database_data.rake
```diff
@@ -0,0 +1,105 @@
...
+    db_host = db_config['production']['host']
+    db_name = db_config['production']['database']
+
+    # Dump data into sql file and gzip
+    `ssh deploy@#{hostname} -p 1855 PGPASSWORD=#{db_password} 'pg_dump --clean \
+                                                                       --no-owner \
+                                                                       -U #{db_username} \
+                                                                       -h #{db_host} \
+                                                                       -d #{db_name} > psql-rmd-prod-data.sql;
+                                                               gzip psql-rmd-prod-data.sql'`
```
rschenk, line 28: I reckon you ought to timestamp the file on the server, just in case two developers happen to run this at the same time. Since this will likely take a few minutes to run, it's not entirely impossible

rschenk: And riffing on the same "this will take a while" note, I do think it would be user friendly to `puts` some status updates along the way, so we can tell that it's doing something. Like maybe even splitting up the pgdump from the gzip, so that you can report the progress

rschenk:
If you really want to go nuts...

```ruby
require 'bundler/setup'
require 'progressbar'

pg_prog = ProgressBar.create(total: nil, title: "pgdump", format: "%t |%B| %a")
Thread.new { until pg_prog.finished? do sleep(0.2); pg_prog.increment; end }
sleep(5) # Do actual pgdump command here
pg_prog.finish

gz_prog = ProgressBar.create(total: nil, title: "gzip", format: "%t |%B| %a")
Thread.new { until gz_prog.finished? do sleep(0.2); gz_prog.increment; end }
sleep(5) # Do actual gzip here
gz_prog.finish

rsync_prog = ProgressBar.create(total: nil, title: "rsync", format: "%t |%B| %a")
Thread.new { until rsync_prog.finished? do sleep(0.2); rsync_prog.increment; end }
sleep(5) # Do actual rsync here
rsync_prog.finish
```

ajkiessl:
That _is_ nuts... I'll give it a try.

awead:
There's yet another way... you can wrap all the command in a Net::SSH session, and either report back the progress results from the server side execution (ie. `rsync --progress`), or do the Thread option, which you would need to do with gzip, since AFAICT, it doesn't have it's own built-in progress option. The details are here:

https://stackoverflow.com/questions/872380/how-do-i-display-progress-bars-from-a-shell-command-over-ssh

``` ruby
Net::SSH.start("host", "user") do |ssh|
  ssh.exec "[command you're running on the server]" do |ch, success|
    bar = ProgressBar.create(total: nil, title: ch, format: "%t |%B| %a")
    until success
      bar.increment && sleep 0.2
    end
    bar.finish
  end

  [repeat above with additional commands]
end
```

This doesn't use ProgressBar, but you could probably wire it in there

ajkiessl:
I'm using the .exec! method from Net::SSH to run the ssh commands.  I also wrapped the time consuming processes in the progress bar code.

awead (approved):

PR #343, opened 2021-11-04
Upgrade RMD to Rails 6.1

### LINKED ISSUE #201: Upgrade to Rails 6
We're currently on Rails 5. 6.1 is the latest, with 7 just around the corner.

## WDLL

* upgrade to Rails 6.0.x or 6.1, whichever is possible with the least amount of hassle

## Dependencies
- [x] #255
- [x] #264
- [x] #287

nicktiberi:
Closes #201.


nicktiberi:
This should be good to go once we figure out why the tests in `open_access_publications/edit_spec.rb` are failing.

There will be some post-Rails 6 tasks once this goes out, including:
- investigating use of https://github.com/rails/jsbundling-rails for bundling assets
- enable the remainder of the Rails 6.1 new framework default settings

Gemfile
```diff
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-gem 'rails', '~> 5.2.4'
+gem 'rails', '~> 6.1.4.1'
```
awead, line 6: I think `~> 6.1` is ok here.

config.ru
```diff
@@ -5,3 +5,4 @@
 require_relative 'config/environment'
 
 run Rails.application
+Rails.application.load_server
```
awead, line 8: I think this line is redundant.

awead (changes requested):
Looks solid to me. I assuming most of this was auto-generated with the built-in update rake task. I had only two minor recommendations.

config.ru
awead: I compared it to Scholarsphere, and it doesn't have that line. 🤷

nicktiberi:
> Looks solid to me. I assuming most of this was auto-generated with the built-in update rake task. I had only two minor recommendations.

Yeah, the vast majority of these changes were done by diffing with the built-in update task and running niftany.

config.ru
nicktiberi: Interesting. The update task recommended adding this. @EricDurante - do you have any thoughts here?

awead: Nm... I checked a `rails new` setup and it is as above with both lines, so leave this as-is.

EricDurante: I'm not sure. If this is indeed incorrect or unnecessary, then it's a mistake in the rails generators/updater. With Rails 6.1, `rails new` generates a rack config file that looks like this.

awead (dismissed):

awead:
I checked the change in coverage and it looks like it's only affecting RailsAdmin::Config::Actions::ExportPublicationsToActivityInsight. But in comparing the two files in CodeClimate, the coverage doesn't change... it's still missing the same amount of lines:

Screen Shot 2021-11-04 at 5 03 09 PM

awead (approved):
Awesome work, @nicktiberi!

PR #348, opened 2021-11-05
Allow deputies to become/unbecome their primaries

### LINKED ISSUE #315: Allow delegate to become a user
A user should be able to sign in as another user that they have been assigned a delegate for.  They should then be able to perform all of the same actions that the user they are a delegate for could.

## WDDLL

* From a list of delegators that have been approved, the delegate can select a user and subsequently log them in as that user
* Use similar devise/current_user logic that is being used for the admin delegation feature
* Banner to remind user that they are logged in as someone else
* Prevent delegate from accessing a their delegator's delegate info
* Set delegate ID field for scholarsphere work deposits, waivers, and open access locations when delegate takes action
* When authenticating each page load, check if delegate status is active so that inactive delegates can't continue
* Delegate should be able to deactivate themselves from being someones delegate

awead:
When a primary user assigns another user as their deputy, they should be able to masquerade as that user and do things on their behalf. This feature already existed for admins, so we're expanding it to include users that are assigned deputy roles from another primary user.

To preserve the existing feature, the original admin functions are extracted to a shared module. The admin controller then only needs to include the core functions of that module. The user controller uses those same features but adds additional protections that ensure only the deputy may become the primary.

Once the deputy is vetted, they become the primary user via CurrentUserBuilder. However, certain circumstances may arise where the primary user removes their deputy, revoking their privileges, but that deputy may still have an active session. To guard against that, we always check the deputies privileges for _each request_. The downside of this is that we can't memoize current_user. CurrentUserBuilder is always called, even if the user it not intending to masquerade as another user.

Fixes #315

### DIFF
--- a/app/controllers/masquerade_controller.rb
+++ b/app/controllers/masquerade_controller.rb
@@ -8,7 +8,7 @@ class MasqueradeController < UserController
   private
 
     def verify_deputies
-      return if primary_user.available_deputies.include?(current_user)
+      return if primary_user.available_deputy?(current_user)
 
       flash[:alert] = I18n.t('profile.errors.not_authorized')
       redirect_to root_path
--- a/app/controllers/masquerade_controller.rb
+++ b/app/controllers/masquerade_controller.rb
@@ -8,7 +8,7 @@ class MasqueradeController < UserController
   private
 
     def verify_deputies
-      return if primary_user.deputies.include?(current_user)
+      return if primary_user.available_deputies.include?(current_user)
 
       flash[:alert] = I18n.t('profile.errors.not_authorized')
       redirect_to root_path
--- a/app/controllers/masquerade_controller.rb
+++ b/app/controllers/masquerade_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class MasqueradeController < UserController
+  include MasqueradingBehaviors
+
+  before_action :verify_deputies
+
+  private
+
+    def verify_deputies
+      return if primary_user.deputies.include?(current_user)
+
+      flash[:alert] = I18n.t('profile.errors.not_authorized')
+      redirect_to root_path
+    end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -142,12 +142,12 @@ class User < ApplicationRecord
     is_active
   end
 
-  def available_deputies
-    self
-      .class
-      .joins(:deputy_assignments)
-      .where(deputy_assignments: { primary_user_id: id, is_active: true, deactivated_at: nil })
-      .where.not(deputy_assignments: { confirmed_at: nil })
+  def available_deputy?(other_user)
+    primary_assignments
+      .where(deputy: other_user)
+      .active
+      .confirmed
+      .any?
   end
 
   def update_psu_identity
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -142,6 +142,14 @@ class User < ApplicationRecord
     is_active
   end
 
+  def available_deputies
+    self
+      .class
+      .joins(:deputy_assignments)
+      .where(deputy_assignments: { primary_user_id: id, is_active: true, deactivated_at: nil })
+      .where.not(deputy_assignments: { confirmed_at: nil })
+  end
+
   def update_psu_identity
     update(psu_identity: psu_identity_data, psu_identity_updated_at: Time.zone.now)
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -64,6 +64,8 @@ Rails.application.routes.draw do
     post 'publications/:id/open_access/waivers' => 'internal_publication_waivers#create', as: :internal_publication_waivers
     get 'publications/open_access_waivers/new' => 'external_publication_waivers#new', as: :new_external_publication_waiver
     post 'publications/open_access_waivers' => 'external_publication_waivers#create', as: :external_publication_waivers
+    post 'unbecome/:user_id' => 'masquerade#unbecome', as: :unbecomes_user
+    post 'become/:user_id' => 'masquerade#become', as: :becomes_user
   end
 
   post 'orcid_access_token' => 'orcid_access_tokens#new', as: :new_orcid_access_token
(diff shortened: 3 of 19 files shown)


ajkiessl (changes requested):

app/controllers/masquerade_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class MasqueradeController < UserController
+  include MasqueradingBehaviors
+
+  before_action :verify_deputies
+
+  private
+
+    def verify_deputies
+      return if primary_user.deputies.include?(current_user)
```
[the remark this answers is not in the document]
awead: Yes, I think so, and probably also in the CurrentUserBuilder as well... We should make a method on user that does this, something like `User#available_deputies`.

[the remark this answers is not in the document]
ajkiessl: Agreed.

app/models/user.rb
```diff
@@ -142,6 +142,14 @@ def active?
     is_active
   end
 
+  def available_deputies
+    self
+      .class
+      .joins(:deputy_assignments)
+      .where(deputy_assignments: { primary_user_id: id, is_active: true, deactivated_at: nil })
+      .where.not(deputy_assignments: { confirmed_at: nil })
+  end
```
rschenk, line 151: I think this should be memoized, but also I think we should write a more efficient version for use with the masquerading. 

Right now we load out all the available deputies, but we only ever ask "is user X an avaliable deputy of user Y" which requires scanning through the list. We could accomplish the same thing faster with an implementation that just hits the table for the pair that we're interested in:

```ruby
Class User
def active_deputy?(another_user)
  deputy_assignments
    .where(deputy: another_user)
    .active
    .confirmed   # This scope will need to be implemented I think
    .any?
end
```

rschenk (changes requested):

app/models/user.rb
awead: I agree on making it more efficient. I'm not sure about the memoization because this has bitten me before. I'm not always sure when the object gets refreshed and a user could be waiting for a new deputy that's been approved but can be used yet because the data is stale.

awead:
```
> primary_assignments.where(deputy: other_user).active.confirmed.explain
=> EXPLAIN for: SELECT "deputy_assignments".* FROM "deputy_assignments" WHERE "deputy_assignments"."primary_user_id" = $1 AND "deputy_assignments"."deputy_user_id" = $2 AND "deputy_assignments"."is_active" = $3 AND (CONFIRMED_AT IS NOT NULL) [["primary_user_id", 7], ["deputy_user_id", 2], ["is_active", true]]
                                                          QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------
 Index Scan using index_deputy_assignments_on_unique_users_if_active on deputy_assignments  (cost=0.15..8.17 rows=1 width=65)
   Index Cond: ((primary_user_id = '7'::bigint) AND (deputy_user_id = '2'::bigint))
   Filter: (is_active AND (confirmed_at IS NOT NULL))
(3 rows)
```

versus

```
> self.class.joins(:deputy_assignments).where(deputy_assignments: { primary_user_id: id, is_active: true, deactivated_at: nil }).where.not(deputy_assignments: { confirmed_at: nil }).explain
=> EXPLAIN for: SELECT "users".* FROM "users" INNER JOIN "deputy_assignments" ON "deputy_assignments"."deputy_user_id" = "users"."id" WHERE "deputy_assignments"."primary_user_id" = $1 AND "deputy_assignments"."is_active" = $2 AND "deputy_assignments"."deactivated_at" IS NULL AND "deputy_assignments"."confirmed_at" IS NOT NULL [["primary_user_id", 7], ["is_active", true]]
                                                            QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.15..9.18 rows=1 width=1003)
   ->  Seq Scan on users  (cost=0.00..1.00 rows=1 width=1003)
   ->  Index Scan using index_deputy_assignments_on_unique_users_if_active on deputy_assignments  (cost=0.15..8.17 rows=1 width=8)
         Index Cond: ((primary_user_id = '7'::bigint) AND (deputy_user_id = users.id))
         Filter: (is_active AND (deactivated_at IS NULL) AND (confirmed_at IS NOT NULL))
(5 rows)
```

rschenk (approved):

awead:
@rschenk I've added a new commit to this because the UI buttons to unbecome/become a user with a deputy were not working. This required some changes to the controller. I went back-and-forth with this from component, to profile decorator/presenter, and back to helper methods. I ultimately went with helper methods because it was already headed that way and I couldn't wrap `profile_for_current_user?` in the component because it's spread out in the view. Ultimately, a decorator/presenter might be in order, but that's another can of works.

awead:
@rschenk also, memoizing `available_deputy?` might be back on the table, but I'm just no sure.

ajkiessl (approved):
This all looks good to me.  Just like with the admins though, we need to come up with a way to direct proxies to the profile of the user they want to become.  I think that's out of the scope of this PR, though.  Just wanted to mention it.

config/routes.rb
```diff
@@ -47,24 +47,29 @@
...
-  post 'profile/publications/:id/open_access/scholarsphere_deposit' => 'open_access_publications#create_scholarsphere_deposit', as: :scholarsphere_deposit
-  get 'profile/publications/:id/open_access/waivers/new' => 'internal_publication_waivers#new', as: :new_internal_publication_waiver
-  post 'profile/publications/:id/open_access/waivers' => 'internal_publication_waivers#create', as: :internal_publication_waivers
-  get 'profile/publications/open_access_waivers/new' => 'external_publication_waivers#new', as: :new_external_publication_waiver
-  post 'profile/publications/open_access_waivers' => 'external_publication_waivers#create', as: :external_publication_waivers
-  post 'orcid_access_token' => 'orcid_access_tokens#new', as: :new_orcid_access_token
-  get 'orcid_access_token' => 'orcid_access_tokens#create', as: :orcid_access_token
 
   get 'profile' => redirect('profile/publications/edit')
+  scope 'profile' do
```
rschenk, line 52: 🔥

rschenk (approved):
Looks good to me!

PR #359, opened 2021-11-10
Rails Admin: remove JS to close side nav

nicktiberi:
Since we upgraded to Rails Admin 2 and removed the Rails Admin Material Theme, the side nav can no longer open and close, so this code no longer does anything.


EricDurante (approved):
Agreed - all dead code should be removed. If write our own feature to allow users to hide the side nav again, then this code would be moved/rewritten anyway.

PR #360, opened 2021-11-11
Allow users to claim publications

EricDurante:
This has the bulk of the functionality for https://github.com/psu-stewardship/researcher-metadata/issues/316, but at this point it only allows searching by publication title, and it doesn't yet send an email notification to an admin when a user claims a publication. The implementation differs slightly from what's described in the issue, but it should be equivalent.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -80,7 +80,7 @@ class Publication < ApplicationRecord
         }
 
   scope :subject_to_open_access_policy, -> { journal_article.published.where('published_on >= ?', Publication::OPEN_ACCESS_POLICY_START) }
-  scope :claimable_by, ->(user) { visible.where.not(id: user.authorships.unclaimable.map { |a| a.publication.id }) }
+  scope :claimable_by, ->(user) { journal_article.visible.where.not(id: user.authorships.unclaimable.map { |a| a.publication.id }) }
 
   scope :open_access, -> { distinct(:id).left_outer_joins(:open_access_locations).where.not(open_access_locations: { publication_id: nil }) }
   scope :scholarsphere_open_access, -> { open_access.where(open_access_locations: { source: Source::SCHOLARSPHERE }) }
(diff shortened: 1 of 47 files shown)


app/models/publication.rb
```diff
@@ -80,6 +80,7 @@ def self.open_access_statuses
         }
 
   scope :subject_to_open_access_policy, -> { journal_article.published.where('published_on >= ?', Publication::OPEN_ACCESS_POLICY_START) }
+  scope :claimable_by, ->(user) { journal_article.visible.where.not(id: user.authorships.unclaimable.map { |a| a.publication.id }) }
```
EricDurante, line 84: Just realized that I ended up with a double negative here. Works fine. Doesn't read that well though..

ajkiessl (dismissed):
I think this all looks good.  I was thinking that rendering a "No publications found" flash or message in the publications search would be helpful for users.  However, I'm not sure it would be possible with the way the controller works.  I don't think it's something that _needs_ to be there, though.

EricDurante:
@ajkiessl Good suggestion! I added that plus a few other things to make the search UI more user-friendly.

app/models/publication.rb
```diff
@@ -80,6 +80,7 @@ def self.open_access_statuses
...
```
rschenk, line 83: Unless authorships automatically preloads its publication, then there's an n+1 here. `a.publication.id` will hit the database for every iteration to load the publication out, simply to ask it for its id, which Authorship already knows. `a.publication_id` could accomplish the same thing without loading out another activerecord

Extra macho factor: `user.authorships.unclaimable.map(&:publication_id)` or `user.authorships.unclaimable.pluck(:publication_id)` depending on whether those authorships are loaded or not.

rschenk (approved):
Looks good, but I did notice one potential `n+1`

ajkiessl (dismissed):

rschenk (approved):

PR #368, opened 2021-11-12
More transactional emails for proxy events

### LINKED ISSUE #367: Proxy Emails: Proxy Revocation and End Proxy Status
In addition to our current slate of proxy-related emails, we should also send emails in the following two scenarios: 

- send an email to the primary user if the proxy decides to end their proxy status
- send an email to the proxy if the primary user revokes the the proxy user's proxy status

nicktiberi:
Closes #367.

- an email is sent to the proxy when the primary user revokes their proxy status
- an email is sent to the primary user when the proxy ends their proxy status

Note that these are not yet hooked up to events. That will come later.


ajkiessl (approved):
Looks good!

PR #371, opened 2021-11-15
335 generate pubs

ajkiessl:
There's a couple of tasks here that generate different works to fill up the profile page.  I also added a couple other generators that generate more complicated things like duplicate groups and non duplicates groups.  I've had to create these to test things in the past and they were a pain to get right in the rails console.


app/models/works_generator.rb
```diff
@@ -0,0 +1,181 @@
...
+        isbn: FFaker::Book.isbn,
+        volume: rand(50),
+        edition: rand(50),
+        doi: "https://doi.org/10.#{rand(1000..9999)}/#{FFaker::Internet.ip_v4_address}",
+        published_on: (Date.today - 6.months)
+      }
+    end
+
+    def rand_journal
+      Journal.count.zero? ? FactoryBot.create(:journal) : Journal.find(rand(Journal.count))
```
rschenk, line 177: There's a chance `Journal.find(rand(Journal.count))` could raise an exception, if a journal is ever deleted then it will leave a "hole" behind in the ids, which will raise a `RecordNotFound` exception if randomly hit. Pretty unlikely but possible.

Another way to do this same without falling into a missing id hole is `Journal.order('RANDOM()').first`

app/models/works_generator.rb
```diff
@@ -0,0 +1,181 @@
+# frozen_string_literal: true
+
+class WorksGenerator
+  def initialize(webaccess_id)
+    raise ArgumentError, 'No Webaccess ID supplied' if webaccess_id.blank?
+
+    @user = User.find_by(webaccess_id: webaccess_id)
```
rschenk, line 7: Maybe this should raise an exception if it can't find that user? Or `find_or_create_by`?

rschenk (commented):
Just two minor details stuck out to me, but I agree with Adam's suggestion to make these as traits in the factories, then we can use them in out tests as well

awead:
@ajkiessl This is look really good. My last comments are to not put anything in app/models. None of this really belongs there anyway. I'd probably put it under `spec/support` since it's so heavily dependent on FactoryBot. The rake task will need to require the file directly as well. Also, if you're wanting to run these tasks in a deployed environment such as staging or qa, you'll need to move the ffaker and factory-bot-rails gems our of the test/dev sections of the gemfile and put them in the top-most group so they'd be deployed to production.

I also might include a guard clause in the WorkGenerator that prevents this from running in a "production production" environment, and only works in dev, test, qa, and staging. I _think_ we're using Rails.env such that we can do this.

ajkiessl:
@awead
>  I'd probably put it under spec/support since it's so heavily dependent on FactoryBot. The rake task will need to require the file directly as well.
I moved `works_generator.rb` to `lib/utilities`.  I'm not sure that WorksGenerator _needs_ to be tested, but since I am testing it, then it might not have made sense putting it in `spec/support`.

rschenk (approved):

PR #373, opened 2021-11-15
Use hi-res versions of the Penn State logo in profile page headers

nicktiberi:
### Before
Screen Shot 2021-11-15 at 2 02 58 PM

### After
Screen Shot 2021-11-15 at 2 03 05 PM


EricDurante (approved):
Looks great!

I don't remember where that original logo came from.

PR #375, opened 2021-11-16
Unify UI of gray title bar for public/manage profile pages

nicktiberi:
We've removed some unnecessary CSS and JS since we are no longer dynamically hiding/showing this bar on the public profile page.

We are now using the same shared CSS to render the gray bar the exact same way on both the public and manage profile pages.

We also stopped importing Open Sans on the public profile page for a few reasons:
- it was causing the font to render slightly differently on the public profile page than on other pages in the app
- we were not importing Open Sans on any other pages in the app (even though we are using Open Sans as a potential font on those other pages)
- removing the import of Open Sans could (slightly) speed up public profile load time

### Public Profile (large)
Screen Shot 2021-11-16 at 12 05 07 PM

### Public Profile (small)
Screen Shot 2021-11-16 at 12 05 13 PM

### Manage Profile
Screen Shot 2021-11-16 at 12 05 34 PM

### DIFF
--- a/app/assets/javascripts/profile/profile_page.js
+++ b/app/assets/javascripts/profile/profile_page.js
@@ -43,19 +43,4 @@ $(document).on('ready', function() {
   $(window).scroll(function() {
     stickyNav();
   });
-
-// Move upper nav to hamburger container
-  var moveNav = function(){
-    if (Foundation.MediaQuery.atLeast('large')) {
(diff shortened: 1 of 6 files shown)


EricDurante (approved):
Love it!

So glad that this UI is getting some needed attention.

PR #376, opened 2021-11-16
332 delegate confirmation

### LINKED ISSUE #332: Confirmation process for delegates
* email is sent to the new delegate for confirmation
 * Delegate must click the confirmation link to sign in to profile where they can confirm their delegacy
* UI elements to reflect status of delegates awaiting confirmation
* UI elements indicating requests that need my approval

rschenk:
Whenever someone does anything with DeputyAssignments, it emails the other party:
* Creation: emails the deputy
* Deputy confirms: emails the primary
* Deputy declines: emails the primary
* Primary revokes: emails the deputy
* Deputy breaks up with primary: emails the primary a youtube to Kelly Clarkson - Since You've Been Gone [looped for 10 hours](https://www.youtube.com/watch?v=83D5T8eue9s)

This also includes some slight changes to the `NewDeputyAssignmentForm` object to create fields that match what `User#attributes_from_psu_identity` does. I did not actually use the latter method because it didn't give me enough control over the error handling. @awead and I are going to refactor both of the aforementioned methods to use a common service class.

Closes #332

### DIFF
--- a/app/mailers/deputy_assignments_mailer.rb
+++ b/app/mailers/deputy_assignments_mailer.rb
@@ -2,9 +2,9 @@
 
 class DeputyAssignmentsMailer < ApplicationMailer
   def deputy_assignment_confirmation(deputy_assignment)
-    @primary = deputy_assignment.primary
+    @primary = UserProfile.new(deputy_assignment.primary)
     @deputy = deputy_assignment.deputy
-    @deputy_url = 'FIXME' # TODO: replace this with the actual URL
+    @deputy_url = deputy_assignments_url
     mail to: @primary.email,
          subject: 'PSU Researcher Metadata Database - Proxy Request Confirmed',
          from: 'openaccess@psu.edu',
@@ -12,7 +12,7 @@ class DeputyAssignmentsMailer < ApplicationMailer
   end
 
   def deputy_assignment_declination(deputy_assignment)
-    @primary = deputy_assignment.primary
+    @primary = UserProfile.new(deputy_assignment.primary)
     @deputy = deputy_assignment.deputy
     mail to: @primary.email,
          subject: 'PSU Researcher Metadata Database - Proxy Request Declined',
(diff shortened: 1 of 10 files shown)


spec/component/mailers/deputy_assignments_mailer_spec.rb
```diff
@@ -3,25 +3,31 @@
...
-                         email: 'primary@psu.edu',
-                         name: 'Primary User',
-                         webaccess_id: 'abc123' }
+  include Rails.application.routes.url_helpers
 
-  let(:deputy) { double 'user',
-                        email: 'deputy@psu.edu',
-                        name: 'Deputy User',
-                        webaccess_id: 'def456' }
+  let(:primary) { instance_double 'User',
```
nicktiberi, line 8: Just curious - what is the benefit here of using `instance_double` over `double`?

app/mailers/deputy_assignments_mailer.rb
```diff
@@ -2,17 +2,17 @@
 
 class DeputyAssignmentsMailer < ApplicationMailer
   def deputy_assignment_confirmation(deputy_assignment)
-    @primary = deputy_assignment.primary
+    @primary = UserProfile.new(deputy_assignment.primary)
```
nicktiberi, line 5: Just a question so I understand - why are we wrapping the primary user in a user profile here, but not the deputy? And in some other instances, wrapping the deputy, but not primary? Is that just so we have access to the `email` field?

nicktiberi (approved):
This looks great! I just have a few questions so I understand things better.

spec/component/mailers/deputy_assignments_mailer_spec.rb
rschenk: `instance_double` ensures that you can only mock real actual methods, whereas `double` allows you to mock any method you want, regardless of whether it exists.

So for example in this very code the `User` model does not actually have a method/field called `email`. When using `double`, it allows this, which can lead to bugs. When I changed it over to `instance_double`, the tests exploded saying "you can't set `email` on `User` because User doesn't have one of those!"

So, tldr, always use `instance_double` or `instance_spy` if you can.

app/mailers/deputy_assignments_mailer.rb
rschenk: Correct, the _only_ thing `UserProfile` does for us is add an `email` method. I could have wrapped them both I suppose.

rschenk:
This entire PR was just an excuse to deduce Alex's guilty pleasure breakup jam

EricDurante:
Is it too late to vote for https://www.youtube.com/watch?v=1NarDEEhOsM?

PR #378, opened 2021-11-16
BaseDecorator maintains `==` with a decorated ActiveRecord model

### LINKED ISSUE #377: A UserDecorator does not equal its User, but should
The following can happen in the the app and can lead to bugs:

```ruby
a_user = User.find(123)
decorated_user = UserDecorator.new(user: a_user)

decorated_user == a_user
# => true

a_user == decorated_user
# => false
```

This is a problem because `current_user` is automatically wrapped by `UserDecorator` and can cause this type of bug in a controller:

```ruby
authorship = current_user.authorships.first

authorship.user == current_user
# => false
```

## WDLL

```ruby
a_user = User.find(123)
decorated_user = UserDecorator.new(user: a_user)

decorated_user == a_user
# => true

a_user == decorated_user
# => true
```

rschenk:
Closes #377

### DIFF
--- a/app/decorators/base_decorator.rb
+++ b/app/decorators/base_decorator.rb
@@ -11,9 +11,9 @@ class BaseDecorator < SimpleDelegator
     __getobj__
   end
 
-  # @note Ensures that type checks return the expected classes. This was causing ActiveRecord.where and .find_by
-  # operations not to work.
-  def is_a?(klass)
-    __getobj__.is_a?(klass)
-  end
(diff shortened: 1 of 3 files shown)


rschenk:
Dear Code Climate:
If this PR's diff coverage is 100%, then why do you think that the total coverage went down by 0.1%? 

Regards,
Schenkleberry


Screen Shot 2021-11-16 at 3 10 56 PM

awead (approved):

PR #380, opened 2021-11-16
Update form hint

EricDurante:
Closes https://github.com/psu-stewardship/researcher-metadata/issues/352

While I was at it, I refactored the similar waiver form views using a combination of shared partials and translations. Since we don't actually have multilingual translations, I didn't put all of the content into the translations file at this point. Instead I just made use of it to eliminate some content duplication including the form field hint that we needed to change.

### DIFF
--- a/app/controllers/external_publication_waivers_controller.rb
+++ b/app/controllers/external_publication_waivers_controller.rb
@@ -19,11 +19,11 @@ class ExternalPublicationWaiversController < ProfileManagementController
   private
 
     def waiver_params
-      params.require(:external_publication_waiver).permit([:publication_title,
-                                                           :journal_title,
-                                                           :reason_for_waiver,
-                                                           :abstract,
-                                                           :doi,
-                                                           :publisher])
+      params.require(:waiver).permit([:publication_title,
+                                      :journal_title,
+                                      :reason_for_waiver,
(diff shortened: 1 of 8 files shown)


rschenk:
So, this works, but you can also accomplish this with SimpleForm's conventions and not need to specify `I18n` calls in the view at all. I did this in the "Create a new proxy" form ([yml file](https://github.com/psu-stewardship/researcher-metadata/blob/ce6e17f65fd6851425a0c11b7be8087e4ca8200a/config/locales/simple_form.en.yml#L13) and [view](https://github.com/psu-stewardship/researcher-metadata/blob/ce6e17f65fd6851425a0c11b7be8087e4ca8200a/app/views/deputy_assignments/index.html.erb#L21)). Not sure if this is desirable but I thought it was worth mentioning.

anaelizabethenriquez (dismissed):
I'm not 100% sure I'm reading this correctly, but I think it looks good. I see the new waiver hint text.

ajkiessl (dismissed):

EricDurante:
@rschenk It took a little bit of fiddling, but you're right - that's a better way to do it. I have a hard time remembering all of the features of all of the things.

rschenk (approved):

PR #381, opened 2021-11-17
Add the deputy user id to related resources

### LINKED ISSUE #338: Add delegated user to other database records
When a delegate use does something on behalf of another user, we need to track that event by adding their identity to that record. For example, a delegate may upload an open access publication into ScholarSphere, but acting as another user. We need to know it was the _delegate_ that did this.

## WDLL

The appropriate database record, either `User` or `DeputyAssignment` should be linked to the following tables:

* ScholarsphereWorkDeposit
* Internal and External Waiver models
* Open Access Location

## Questions

1. We need to know whether to use the actual User record or the DeputyAssignment record.

Answer: Use the user id. Based on some forthcoming changes based on the UI implementation, primary/deputy links won't be deleted if they have existing publications so we can rely on the user id and link it back to the primary.

Note: If we want to track administrators, we have to use the user id because admin can't have deputy assignments.

awead:
The first commit adds the deputy user id to the related database tables via an included module. The Rails relation only does the belongs_to side of it, and not the has_many/has_one. This is for simplicity's sake and because we don't need the reverse associations.

The second commit adds the deputy's id to the different controllers that create the resource with associated deputy user ids. The deputy user id is always added as a parameter, but if the current user isn't acting as a deputy, the id will be nil.

Fixes #338

### DIFF
--- a/app/models/concerns/deputy_user.rb
+++ b/app/models/concerns/deputy_user.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module DeputyUser
+  extend ActiveSupport::Concern
+
+  included do
+    # rubocop:disable Rails/InverseOf
+    belongs_to :deputy,
+               class_name: 'User',
+               foreign_key: :deputy_user_id,
+               optional: true
+    # rubocop:enable Rails/InverseOf
+
+    validate :deputy_must_be_assigned
+  end
+
+  private
+
+    # @note We can't tell (easily) if the deputy is associated with the exact user, but we can tell if they have any
+    # valid primary users.
+    def deputy_must_be_assigned
+      return if deputy.nil?
+
+      errors.add(:deputy, :not_assigned) if deputy.deputy_assignments.limit(1).empty?
+    end
+end
--- a/app/controllers/external_publication_waivers_controller.rb
+++ b/app/controllers/external_publication_waivers_controller.rb
@@ -6,7 +6,9 @@ class ExternalPublicationWaiversController < ProfileManagementController
   end
 
   def create
-    @waiver = current_user.external_publication_waivers.build(waiver_params)
+    @waiver = current_user
+      .external_publication_waivers
+      .build(waiver_params.merge(deputy_user_id: current_user.deputy.id))
     @waiver.save!
     flash[:notice] = I18n.t('profile.external_publication_waivers.create.success')
     FacultyConfirmationsMailer.open_access_waiver_confirmation(UserProfile.new(current_user), @waiver).deliver_now
--- a/app/controllers/internal_publication_waivers_controller.rb
+++ b/app/controllers/internal_publication_waivers_controller.rb
@@ -8,7 +8,7 @@ class InternalPublicationWaiversController < OpenAccessWorkflowController
 
   def create
     authorship = current_user.confirmed_authorships.find_by!(publication_id: params[:id])
-    waiver = InternalPublicationWaiver.new(waiver_params)
+    waiver = InternalPublicationWaiver.new(waiver_params.merge(deputy_user_id: current_user.deputy.id))
     waiver.authorship = authorship
     waiver.save!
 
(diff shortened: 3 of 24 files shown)


rschenk:
This looks pretty clean to me. My only comments are pretty minor: You could use [yml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to DRY up the en.yml file, and also that I have an irrational hatred for those `# rubocop:disable` comments in line with the code.

app/models/concerns/deputy_user.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+    # @note We can't tell (easily) if the deputy is associated with the exact user, but we can tell if they have any
+    # valid primary users.
+    def deputy_must_be_assigned
+      return if deputy.nil?
+
+      errors.add(:deputy, :not_assigned) if deputy.deputy_assignments.limit(1).empty?
```
EricDurante, line 24: Is this better than `deputy.deputy_assignments.none?`?

app/views/open_access_publications/edit.html.erb
```diff
@@ -106,6 +106,7 @@
                         as: :boolean,
                         required: true,
                         label: "I have read and agree to #{link_to 'the deposit agreement', 'https://scholarsphere.psu.edu/agreement', target: '_blank'}.".html_safe %>
+            <%= f.hidden_field :deputy_user_id, value: current_user.deputy.id %>
```
EricDurante, line 109: How come we pass the deputy ID through the web form here? Can't we retrieve and save this value entirely on the backend? We don't need to worry much about it being manipulated in this case, but the fact that it _can_ be manipulated smells to me. I don't think that this should be input from the client, even if it's hidden from the user in the UI.

awead:
@rschenk I also have a dislike for inline rubcop disable comments, but it was the only way to get to work here. Disabling it for the one file seemed to cause it to complain elsewhere. I'll take a look again and see if I can't get around that.

app/models/concerns/deputy_user.rb
awead: I'll take a look at the sql that's generated. I don't have a preference for either, but `none?` does read better.

app/views/open_access_publications/edit.html.erb
awead: I supposed we can just merge the id in the params object in the controller?

EricDurante: Yeah, I think that's a much better way to go.

app/models/concerns/deputy_user.rb
awead: In looking at the logs, calling `none?` instead of `limit(1)` produces NO sql output. It works, but I'm not seeing any indication that it's looking at the database. So I gather it must be using some kind of cached reference? Which I guess is ok? 🤷

EricDurante: Interesting. I don't have a strong preference here. I was just curious.

EricDurante (approved):
Looks good!

app/models/concerns/deputy_user.rb
rschenk: `.none?` is the better option here. Ultimately the SQL produced is very similar: 

```sql
# deputy.deputy_assignments.none?
SELECT 1 AS one FROM "deputy_assignments" WHERE "deputy_assignments"."deputy_user_id" = $1 LIMIT $2
```

but as you can see from the `SELECT 1`, AR will not actually load anything out of the database or instantiate an AR object. That's negligibly more efficient (probably doesn't matter) and more readable (does matter)

rschenk:
> Disabling it for the one file seemed to cause it to complain elsewhere.

@awead This happened to me before in this project and it had to do with `.rubocop.yml` vs `.rubocop_todo.yml`

PR #385, opened 2021-11-17
AI Importer: import the 'ACCESIBLE' field from AI publications

### LINKED ISSUE #189: Add 'accessible' field to publication records that pull from AI.
@nmg110 can provide additional details for this, but pretty sure she wants a field that pulls in a status for AI records.

nicktiberi:
Closes #189.

We import the `ACCESIBLE` field from publication data coming from Activity Insight, and store it on the publication model in RMD as `activity_insight_postprint_status`.

This new field is now filterable on the publication list page in Rails Admin.

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -154,7 +154,7 @@ class ActivityInsightImporter
             pub_record = pi.publication
 
             if pi.persisted?
-              pub_record.update!(pub_attrs(pub)) if pub_record.updated_by_user_at.blank?
+              update_pub_record(pub_record, pub)
             else
               pi.save!
             end
@@ -202,6 +202,22 @@ class ActivityInsightImporter
   end
 
   private
+
+    def update_pub_record(pub_record, ai_pub)
+      if pub_record.updated_by_user_at.blank?
+        pub_record.update!(pub_attrs(ai_pub))
+      else
+        # If the publication has been updated by an admin, we only want to
+        # overwrite a subset of its attributes.
+        pub_record.update!(always_update_pub_attrs(ai_pub))
+      end
+    end
+
+    def always_update_pub_attrs(pub)
+      {
+        activity_insight_postprint_status: pub.activity_insight_postprint_status
+      }
+    end
 
     def pub_attrs(pub)
       {
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -917,6 +917,8 @@ class ActivityInsightPublicationPostprint
   end
 
   def postprint_status
+    # NOTE: this field is misspelled as "ACCESIBLE" (one S) on the Activity Insight side,
+    # so the below misspelling is intentional.
     text_for('ACCESIBLE')
   end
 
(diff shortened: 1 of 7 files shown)


EricDurante (commented):
This looks good to me, but I have a couple of possible suggestions.

First, I'm assuming that at some point in the future we will have logic in RMD that depends on the value of this new field. With that being the case, and since we know all of the possible values for `activity_insight_postprint_status`, I think that we should add a validation on the publication model to ensure that the incoming value is one that we know how to handle. That way if new possible values are introduced in Activity Insight, our importer will start throwing validation errors and alert us to the fact that there's a new value that our logic may need to take into account. Otherwise, it would be easy to passively accept and ignore new values that could actually have some significance to us without even being aware of it.

Second, since this is a new publication attribute that is specific to Activity Insight, and since we expect the value to change and need to know when it does, I would make an exception to our rules about how this attribute is updated by the importer. For all of the other publication attributes, we stop allowing the importer to update them once the record has been edited by an admin. I think we want to allow the importer to continue to update this attribute even after an admin has edited the record.

nicktiberi:
I implemented the feedback from @EricDurante and this should be ready for re-review.

app/importers/activity_insight_importer.rb
```diff
@@ -883,3 +910,21 @@ def for_external_person?
...
     end
 end
+
+class ActivityInsightPublicationPostprint
+  def initialize(parsed_postprint)
+    @parsed_postprint = parsed_postprint
+  end
+
+  def postprint_status
+    text_for('ACCESIBLE')
```
awead, line 920: @nicktiberi Not to be that annoying over-the-shoulder person, but shouldn't it be 'ACCESSIBLE' ?

nicktiberi: Yes it should be. 🙂 It's just misspelled as "ACCESIBLE" coming from the Activity Insight end, so that's what we have to look for in the XML.

awead: I'm literally crying right now.

awead:
``` ruby
class Accessible
  def to_s
    'ACCESIBLE'
  end
end
```

nicktiberi: You're not alone 😁 Ideally this would be fixed on the AI side, but it is what it is for now, I guess.

awead: Sorry, I'm crushed by the weight of the irony.

ajkiessl (dismissed):

nicktiberi:
@ajkiessl - I just added a note to clarify why "ACCESIBLE" is misspelled in the importer code. When you get a chance, could you re-approve? Thanks!

app/importers/activity_insight_importer.rb
```diff
@@ -154,7 +154,7 @@ def call
             pub_record = pi.publication
 
             if pi.persisted?
-              pub_record.update!(pub_attrs(pub)) if pub_record.updated_by_user_at.blank?
+              update_pub_record(pub_record, pub)
             else
```
EricDurante, line 158: I'd argue that this logic (both what was here previously and the new changes) actually belong in the `Publication` model since we're asking the publication something about its state and then telling it do one thing or another depending on the answer ([Tell Don't Ask](https://www.martinfowler.com/bliki/TellDontAsk.html)). Something like `pub_record.update_from_ai_import(ai_pub)` could maybe contain all of this logic. There's no need to do anything about this now, but it might be a good refactoring for the future, especially since there are probably similar situations in some of the other importers.

PR #391, opened 2021-11-18
Miscellaneous Rubocop Fixes

nicktiberi:
I went through our `.rubocop_todo.yml` file and addressed some of the low-hanging fruit that it was complaining about.

I put each of these fixes in a separate commit, so it would probably be helpful to review this PR commit-by-commit.


rschenk (approved):
Looks good to me, thanks for doing this

PR #397, opened 2021-11-23
Adds link to primary's public profile on proxies page if confirmed.

ajkiessl:
Added some help text too.  I think the tests I added should be sufficient.

### DIFF
--- a/app/components/deputy_assignment_component.html.erb
+++ b/app/components/deputy_assignment_component.html.erb
@@ -1,10 +1,6 @@
 <div id="<%= dom_id deputy_assignment %>" class="<%= root_class %>">
   <div class="deputy-assignment__name">
-    <% if current_user_is_deputy? && !pending? %>
-      <%= link_to name, profile_path(primary.webaccess_id) %> <%= webaccess_id %>
-    <% else %>
-      <strong><%= name %></strong> <%= webaccess_id %>
-    <% end %>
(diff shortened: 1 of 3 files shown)


app/components/deputy_assignment_component.html.erb
```diff
@@ -1,6 +1,10 @@
 <div id="<%= dom_id deputy_assignment %>" class="<%= root_class %>">
   <div class="deputy-assignment__name">
-    <strong><%= name %></strong> <%= webaccess_id %>
+    <% if current_user_is_deputy? && !pending? %>
```
rschenk, line 3: Just wanted to let you know that there is a method for this: `confirmed?`  

`confirmed? == !pending?` (I personally dislike `!`s in conditionals but that's just me)

app/components/deputy_assignment_component.html.erb
```diff
@@ -1,6 +1,10 @@
...
+      <%= link_to name, profile_path(primary.webaccess_id) %> <%= webaccess_id %>
+    <% else %>
+      <strong><%= name %></strong> <%= webaccess_id %>
```
rschenk, line 6: This entire block could also be done as...

```
<strong><%= link_to_if (current_user_is_deputy? && confirmed?), name, profile_path(primary.webaccess_id) %></strong> <%= webaccess_id %>
```

It's a little more DRY but also a little more terse. Just thought I'd throw it out there.

rschenk (dismissed):

Issue #388, opened 2021-11-18
Refresh user identity information with PSU Identity API


anaelizabethenriquez:
Identity information for existing RMD users, including names, should be refreshed periodically from the Penn State Identity API. Ideally we should refresh this pretty frequently, so that we're not using out-of-date names in public profiles or OA reminder emails.

Users make these changes at https://accounts.psu.edu/manage. We can refer them there and ensure we're refreshing data frequently to get those changes. We can also make changes to their name in RMD.

ajkiessl:
When implementing this, move `User.update_psu_identity` to the `PsuIdentityUserService` class.  The `update_psu_identity` method probably doesn't belong in the `User` if we have a service and we are doing bulk updates.

ajkiessl:
This is something that should be run before OA email notifications go out.

ajkiessl:
Since we will be running the OA email task weekly we should run this at least weekly.

ajkiessl:
@anaelizabethenriquez In the case that an admin has edited a User record (potentially editing the User's name), do we still want the importer to overwrite these changes?  I think this would be preferred, since we would like identity services to be the central authority on names.

anaelizabethenriquez:
@ajkiessl Yes, I agree. I think the importer should overwrite changes.

ajkiessl:
Cron jobs are configured here: https://github.com/psu-libraries/researcher-metadata-chef/blob/f625fafd7636baa0cdcaaf472d6332e0821c1c72/cookbooks/psu_research_metadata/recipes/production.rb

Issue #364, opened 2021-11-11
Fix sign-in so that it is not dependent on identity services


ajkiessl:
We need to catch the error that psu_identity will throw if there is a network error.

ajkiessl:
Blocked by: https://github.com/psu-libraries/psu_identity/issues/12

rschenk:
This should be done in conjunction with #382

anaelizabethenriquez:
Per @ajkiessl, this should be considered for potential development this round. I've moved it up the list on the project board.

PR #398, opened 2021-11-23
Refactor `User.from_omniauth` and `NewDeputyAssignmentForm` to use a common PsuIdentity service object

### LINKED ISSUE #382: Refactor PSU Identity into its own service object
The code that creates a User from OAuth and the code that invites a new proxy, share common functionality and should be refactored to use a common service object. 

I can see this object having two methods:
* Given a webaccess id, find or initialize a user. Returns the user if it finds one in the database, and initializes (but doesn't save) a User with values from the PSU Identity server otherwise
* Given an existing User, fetch data from PSU Identity, and update the fields on that User with these fresh values (personally I don't think this method should save the record to the db, just update the AR record's fields in memory).

The first method should handle the case of webaccess id not being found in the identity server (returning `nil` works) and both methods should handle network outages and other exceptions by raising custom errors that can be rescued appropriately.

rschenk:
This works but is semi-complete. This PR adds a service object that will look in the database for a User with a certain webaccess id, if none exists then it will query PSU's identity server and initialize a User with the results.

Both `User.from_omniauth` and `NewDeputyAssignmentForm` share this functionality.

There's still a little work to be done before I would consider this to be fully complete, but seeing as how all of this additional work is kinda in Adam's wheelhouse, I didn't want to tackle it without getting his feedback first:
* `User.from_omniauth` needs to handle the following errors conditions (which it currently does not and never did):
  * When PSU Identity returns `nil` (no user found)
  * When PSU Identity is down (raises IdentityServiceError) Related to #364 
  * When PSU Identity returns a successful response, but that response is deemed invalid by the User model (first or last name is blank)
* `User.update_psu_identity` should also probably use this service class, which will require an additional public method to be added. Frankly I'm not entirely sure what this method does and why (i.e. why does it update the `psu_identity` hash but not the first/last name?) so maybe Adam is better suited to tackle that. Related to #388 

Closes #382

### DIFF
--- a/app/forms/new_deputy_assignment_form.rb
+++ b/app/forms/new_deputy_assignment_form.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class NewDeputyAssignmentForm
-  class IdentityServiceError < StandardError; end
-
   include ActiveModel::Model
 
   attr_accessor :primary,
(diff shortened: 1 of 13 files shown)


awead:
@rschenk I'll reply to each point above:

`User.from_omniauth`
1. I can't think of a legitimate reason we'd get no user. We should only be calling this method when they're logging into the application, and they wouldn't be able to do that without first going through Azure. It's seems impossible for someone to be authenticated via Azure and _not_ be in PSU's identity system. But, that's an assumption. If it were to happen, that seems to be a "WHAAA!!???!!!" kinda moment. Maybe throw something in Bugsnag and return to the root page with "Sorry, we can't log you in"
2. We can probably defer this to after #364 is completed, then the service can raise its error, and we do the Bugsnag/redirect to root dance.
3. This seems to be more legitimate than the first case, but if we've made our User model incongruent with Penn State's own identity data, that's a shame-on-us moment. Otherwise, it seems unlikely that an Azure-authenticated user wouldn't have both first and last names. Either way, I think the response would the same: "WHAAA??!!??"

`User.update_psu_identity `
Yeah, I think this can just be moved to the service class. `User` doesn't need to do this. The process for refreshing the data can be done in #388 and it'll just call the appropriate class method on the service.

awead (approved):
I don't think there's anything missing from this PR that would block it. There are additional edge cases that need to be addressed, but I don't we're introducing any regressions. It'd be nice to have the `from_omniauth` edges handled, and to have an update method on the service class, but I don't think that's presently required to get this merged. So I would vote to ticket those remaining issues and move on.

Issue #399, opened 2021-11-23
Admin cannot "unbecome" user


anaelizabethenriquez:
While testing in QA, I noticed that as an admin I cannot "unbecome" a user. I get the error message "Your account is not authorized to perform this action." I have been using this feature in production, where it is working fine.

Update: This is only a problem when I click on the "Unbecome this user" button. If I click "stop being [username]" in the header, it works fine.

ajkiessl:
@anaelizabethenriquez There was a bug causing this.  I'm working on a patch to fix it.

PR #401, opened 2021-11-24
Fixes for when someone is masquerading unbecomes a user

ajkiessl:
@awead Can you check this out?

There was an error popping up when and admin or proxy clicked "Unbecome the User".  To fix this, I added some logic that takes into account that when someone is masquerading, the impersonator should be used when checking if current_user is an admin, or one of the primary's available deputies.

### DIFF
--- a/app/controllers/masquerade_controller.rb
+++ b/app/controllers/masquerade_controller.rb
@@ -8,7 +8,11 @@ class MasqueradeController < UserController
   private
 
     def verify_deputies
-      return if primary_user.available_deputy?(current_user)
+      if current_user.masquerading?
+        return if primary_user.available_deputy?(current_user.impersonator)
+      else
+        return if primary_user.available_deputy?(current_user)
+      end
 
       flash[:alert] = I18n.t('profile.errors.not_authorized')
       redirect_to root_path
--- a/app/controllers/masquerade_controller.rb
+++ b/app/controllers/masquerade_controller.rb
@@ -10,8 +10,8 @@ class MasqueradeController < UserController
     def verify_deputies
       if current_user.masquerading?
         return if primary_user.available_deputy?(current_user.impersonator)
-      else
-        return if primary_user.available_deputy?(current_user)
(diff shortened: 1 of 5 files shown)


ajkiessl:
closes #399

app/controllers/masquerade_controller.rb
```diff
@@ -8,7 +8,11 @@ class MasqueradeController < UserController
   private
 
     def verify_deputies
-      return if primary_user.available_deputy?(current_user)
+      if current_user.masquerading?
+        return if primary_user.available_deputy?(current_user.impersonator)
+      elsif primary_user.available_deputy?(current_user)
+        return
+      end
```
awead, line 15: I might do something like:

``` ruby
def current_deputy
  if current_user.masquerading?
    current_user.impersonator
  else
    current_user
end
```

Then the change is just: `return if primary_user.available_deputy?(current_deputy)`

I'm also wondering if this logic should be in `UserDecorator` like:

``` ruby
def as_deputy
  return self unless masquerading?

  self.impersonator
end
```

app/controllers/profiles_controller.rb
```diff
@@ -64,7 +64,7 @@ def path_to_become_user
     end
 
     def path_to_unbecome_user
-      if current_user.admin?
+      if current_user.impersonator.admin?
```
awead, line 67: I believe with the added change to UserDecorator as above, this would read `current_user.as_deputy.admin?` which may or may not make more sense....

awead (approved):
This is fine, but I'm pointing out some clarifications. The deputy logic is getting spread out and it should be contained. I know we're out of time for the sprint, so if you want to leave this as-is, go a head, but I would write a ticket that refactors the current_user/deputy distinction into one location. I'm positive we'll be running into this issue again later where we need to know who the actual deputy is.

ajkiessl:
@awead Agreed.  Thanks for looking at this.  I'll add tickets for those refactorings.

PR #409, opened 2021-12-01
Removed validation that deputy user must have at least one deputy ass…

ajkiessl:
…ignment.  Removed tests for this.


awead (changes requested):
The shared spec is still needed to check that we can actually assign any user to the field. Also, I think you should be able to ignore the VCR changes.

ajkiessl:
I'm going to keep the vcr changes.  I think they need to be added, but I might be missing some context around why they weren't there.

ajkiessl:
Or maybe I'll remove them.  I haven't really used vcr.  Do you only need them if you are referencing them in a test?

awead:
I think the VCR changes aren't relevant to the PR. We could have a mis-configured VCR setup and the yaml files are being changed when the tests are run, which shouldn't happen. If you remove the yaml changes, CI should still pass.

PR #412, opened 2021-12-06
22 deduplicate on doi

### LINKED ISSUE #22: De-duplicate Publications
Conversation on basecamp - https://basecamp.com/1759334/projects/15421427/messages/88508508#comment_732991916 and https://basecamp.com/1759334/projects/15421427/messages/89238237#comment_741254712

We are working around getting publications deduplicated before turning on DM pulling from publications.

### LINKED ISSUE #374: Merge publications based on DOI
We've discussed merging publications that have the same DOI, without requiring manual review. I thought this was in place already, because @ajkiessl ran a task this morning to automatically deduplicate some publications. However, I'm seeing the following duplicate groups where the publications have matching DOIs (I won't merge these for now):

- https://metadata.libraries.psu.edu/admin/duplicate_publication_group/101835
- https://metadata.libraries.psu.edu/admin/duplicate_publication_group/101840
- https://metadata.libraries.psu.edu/admin/duplicate_publication_group/101848
- https://metadata.libraries.psu.edu/admin/duplicate_publication_group/101861 (only 2 out of 3 have DOIs, but they match)
- https://metadata.libraries.psu.edu/admin/duplicate_publication_group/102526

Seems like there are still a lot of these.

ajkiessl:
Adds another auto merge method for duplicate groups that merges publications with the same doi (if they are in the same duplicate group).  The difficulty with this was implementing logic to choose which attributes are preferred between two publications to be merged.  This uses a lot of conditional logic, so it's a bit complicated and difficult to test.  I tried to cover as many variations of inputs as possible in the tests.  The tests are pretty extensive.  I also make an attempt to choose the preferred contributor names between the two publications to be merged. 

Closes #22 
Closes #374


ajkiessl:
@EricDurante If you get some time can you review this PR?  This is the last straggling thing I had to get done for RMD.  It's a bit of a doozy.  It works well, but I'm no overall very pleased with how complex it got (especially since this is something we want to modify in the future).  I'm not sure there's another way to go about it though.

spec/component/models/contributor_name_merge_policy_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+          end
+        end
+      end
+
+      context 'when contributors have a different last name' do
+        context 'when contributors have the same position' do
+          let(:contributor1) { ContributorName.create first_name: 'Test', last_name: 'Tester', position: 1 }
+          let(:contributor2) { ContributorName.create first_name: 'Test', last_name: 'Terry', position: 1 }
+
+          it 'treats these contributors as the different' do
```
EricDurante, line 92: @ajkiessl I'll post more comments, but here's one observation:

I've seen real-world examples where I think this behavior wouldn't be what we want. We sometimes think of a person's name as an immutable (if not unique) identifier for the person, but unfortunately it's not. I've seen cases where it looked like a person was listed as a contributor under their original last name in one set of publication metadata and under the last name that they took when they got married in a duplicate set of the publication metadata. Same person, same position, but different last name. Ideally we'd only be getting metadata that lists peoples' names as they were actually printed in the publication text, but I know that this isn't always the case.

So it looks like we'd end up listing two different names for the same person in cases like that. Does that sound right to you?

While it's less common, people do change their first names for various reasons too. So the same thing could conceivably happen with the first name comparison, though I haven't seen an actual example of this yet.

This is one reason why I started to tie ContributorName records to User records whenever possible. That could allow us to look up a User by ContributorName metadata and see any other names under which they may have published research. That could help us to determine whether or not two contributors with different names are actually the same person. Unfortunately, it can be a chicken and egg problem in terms of getting that data to begin with. At some point, someone has to tell us that it's actually the same person with different names.

So unfortunately I was never able to come up with a decent solution for this problem..

EricDurante:
@ajkiessl This is a really tough problem - both in terms of deciding how it should behave and also in terms of deciding how to design the implementation.

With regard to behavior, I think what you're doing here looks pretty sensible. By now you've definitely given much deeper consideration to most of the details than I have, but overall I think it looks reasonably conservative in terms of what it's attempting to do at this point.

With regard to the design of the code, I can definitely relate to what you're saying about the complexity of this logic. I don't think that there's any way to reduce the inherent complexity of the solution to this problem without making it a worse solution. In fact, we may find that it actually needs to be even more complex than this to do the job well. So we might just have to ask the question again, "Is this actually a problem worth tackling?" If we still decide that the answer is "yes", then we're stuck with the inherent complexity, and all that we can do is try to figure out the best way to organize the labyrinthian logic. As is often the case when ActiveRecord is involved, the persistence framework can make this kind of awkward no matter what we do. The only tool that I have for this kind of problem is breaking the logic down into more small, independent units and adding in more layers of abstraction. However, at least to me, that doesn't feel like an especially natural way to handle this kind of complexity. So I honestly don't really have a clear sense as to the best way to refactor stuff like this in good OOP style (or any other style for that matter). Maybe someone with a better head for design than I have can teach me something here.

spec/component/models/contributor_name_merge_policy_spec.rb
ajkiessl: @EricDurante That sounds right to me.  I'll add something to the ContributorNameMergePolicy that keeps unique contributor names on their `user_id` (if present) after all the filtering.  That way if we have Jane Smith and Jane Johnson that are both linked to the same `User` record, we'd know these are the same person and can get rid of one.  That could help fix some of the instances above.

I'm not sure there's any way right now to handle cases when we don't have a `user_id`, though.

EricDurante: Yeah, there may not be. Perhaps there will be few enough instances of this that it won't be a show-stopper. Probably the worst case scenario at this point is that people who are using our API to populate websites with researcher/publication data will end up displaying one of these mistakes every now and then until someone notices it and we can correct it. It won't look great, but it might be pretty rare.

In the case where we do have a Jane Smith and a Jane Johnson who turn out to be the same person and we need to decide which name to keep, we might just have to go with the more reliable data source (if there is one).

PR #416, opened 2021-12-13
.uniq the query array when querying api on Activity Insight ID and DOI

ajkiessl:
Sending `#publications` to an instance of `APIToken` will return duplicate records if a user is in multiple organizations.  I'm `uniq`ing the array returned from `Publication.filter_by_activity_insight_id` and `Publication.filter_by_doi` so these duplicates aren't returned by the API.  I added some tests for this too.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -584,7 +584,7 @@ class Publication < ApplicationRecord
       .where(publication_imports: {
                source: 'Activity Insight',
                source_identifier: activity_insight_id
-             })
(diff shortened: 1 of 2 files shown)


ajkiessl:
@nicktiberi Can you take a glance at this?

nicktiberi (approved):
LGTM! 👍 This was not something I had considered when I wrote those queries.

ajkiessl:
No problem, thanks!

PR #438, opened 2022-02-15
Fix bug in ORCiD integration by switching to using new method name

EricDurante:
I noticed an error in Bugsnag related to someone trying to add employment info to ORCiD. It turned out to be a regression caused by the last Rails upgrade that we didn't catch because our tests in this area are a little thin. I went ahead and made the fix since it was quick, but I haven't done anything to improve the associated tests. That might be an item for future consideration.

### DIFF
--- a/app/controllers/orcid_employments_controller.rb
+++ b/app/controllers/orcid_employments_controller.rb
@@ -10,7 +10,7 @@ class OrcidEmploymentsController < UserController
       else
         employment = OrcidEmployment.new(membership)
         employment.save!
-        membership.update_attributes!(orcid_resource_identifier: employment.location)
+        membership.update!(orcid_resource_identifier: employment.location)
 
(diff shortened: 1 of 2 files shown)


ajkiessl (approved):

anaelizabethenriquez:
Thank you both! I'm not sure if there's a way for us to delete anything in ORCID for a user, but @DanCoughlin or @bdezray might know.

anaelizabethenriquez:
@DanCoughlin @bdezray Do you know whether we can fix the user's ORCID record for them? If not, I would like to email them.

DanCoughlin:
@anaelizabethenriquez I apologize for not responding sooner. I considered Eric's answer to be more informative than what I had. I should have confirmed that for you.

anaelizabethenriquez:
@DanCoughlin Sounds good, thanks! I will email the user.

PR #485, opened 2022-08-03
Upgrade to ruby 3.1.2

ajkiessl:
Updates to ruby 3

Some notable things:

- Removed mini_racer.  It was not being implemented anywhere, so I'm assuming it wasn't used.  It's system packages were difficult to maintain
- Net-smtp, net-imap, and net-pop must be specified in the Gemfile for ruby3 to work with Rails 6.   Also, strscan should be set at 3.0.1 to work with bundler 2.3.8
- net-ssh@7.0.0.beta1 is needed for capistrano to work
- Switched from unmaintained fast_jsonapi to jsonapi-serializer for ruby Object serialization
- Updated apivore to work with ruby 3: https://github.com/psu-libraries/apivore/pull/1
- Updated the chef code to configure server properly: https://github.com/psu-libraries/researcher-metadata-chef/pull/4


ajkiessl:
Deploy notes:

- Run at a time when the chef cron won't interfere (run at the 18th minute of every hour)
- Run the chef code first to get the proper ruby version installed
- It will likely fail when it gets to the delayed job stage since it won't have the proper gems installed
- Then deploy with capistrano
- Finally, run the chef code again just to check that there are not errors

PR #497, opened 2022-08-16
Activity Insight import error catching

### LINKED ISSUE #413: AI IDs associated with 0 or multiple RMD records
@srerickson has been querying the RMD API by AI ID and gotten some unexpected results. Quite a few of the IDs he's looking for have either zero records in RMD or more than one record (up to 12). Seems like this might have something to do with duplicates, but I'm not sure exactly what. This is not blocking our work, it's just very unexpected.

These AI IDs are coming from Activity Insight exports that @nmg110 's team runs for publications where the researcher uploads a post-print in RMD. They can have a variety of document types in AI (not necessarily journal articles). 

Here are examples of AI IDs with no records associated:

222649597952
218823843840
214945056768
218861940736
219108661248

Here are examples of AI IDs with more than one record associated:

196613644288 (3 records)
203850590208 (2 records)
214769737728 (12 records)
214769721344 (12 records)
218180702208 (11 records)

### LINKED ISSUE #417: Fine-tune error handling in ActivityInsightImporter
Currently the scope of error catching in `ActivityInsightImporter` is too broad.  The import iterates through a list of users obtained from the Activity Insight API.  It then uses the API to grab all the data for this user.  If there is an error when updating/creating any records associated with a user (Presentations, publications, contributors, etc.) the error will be caught and the importer will move to the next user.  For example, if a publication is being imported that doesn't have a title, a validation error will be raised, rescued, and anything after that for that user will be skipped.  It just moves to the next user.

There needs to be a `rescue` block for each section of the `ActivityInsightImporter` (Education history, presentations, publications, etc.).  The `rescue` block should then report the error and record metadata to the `ImporterErrorLog`.  It should be noted, though, that there may be *__a lot__* of errors.

ajkiessl:
ActivityInsightImporter changes to catch errors for every record for every major import type (User, EducationHistoryItem, Presentation, Performance, Publication).  Information about the caught errors are then logged in the ImporterErrorLogs.  This information includes the AI model that was affected, the webaccess id of the User that the import was for, and little metadata about the record (ID in AI, and timestamps).

closes #417 
closes #413


ajkiessl:
This is ready for review, but I want to wait until the import is done running in QA before merging.

banukutlu (approved):
👌🏻 🚢

PR #499, opened 2022-08-17
Open access url validation fix

### LINKED ISSUE #459: Open access URL validation may not be working correctly in some cases again
See https://github.com/psu-stewardship/researcher-metadata/commit/f3a8476285d409babb6b083835c10085f6b7f888

### LINKED ISSUE #487: Validation on user-submitted OA URL fails even though URL appears good
Both the user and I have tried entering the following URL, over several days. The URL resolves fine for me in my browser, but RMD gives the following error: "URL is invalid or is unavailable at this time."

URL: https://www.researchgate.net/publication/358621860_Examining_the_Link_Between_Neutral_and_Ambivalent_Attitudes_Their_Association_and_their_Co-Occurrence

banukutlu:
Closes #459

### DIFF
--- a/spec/component/forms/open_access_url_form_spec.rb
+++ b/spec/component/forms/open_access_url_form_spec.rb
@@ -118,20 +118,12 @@ describe OpenAccessURLForm do
   end
 
   # One brittle test for a real-world edge case
-
-  # TODO:  Something about this stopped working at some point. The request that
-  # we're making for this URL with HTTParty is now returning 503 instead of 200.
-  # If I make what looks like the exact same request with curl (using the same user
-  # agent header), I still get 200. I'm not sure why this is happening, but it
-  # probably means that the URL validation has stopped working correctly in some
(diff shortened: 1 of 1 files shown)


ajkiessl (commented):
I think we should either remove this or rewrite it: https://github.com/psu-libraries/researcher-metadata/pull/499/files#diff-ab404598a37eb4535dc6aafe41f4cd23a3e3b7ab3b16dff77fc45e8dcd454dd6R127-R132

Since it seems to be working now.

Also, we should probably add a test in the mocked portion of the tests in that file that tests when the first request fails but the second one succeeds.

banukutlu:
@ajkiessl I took my user-agent changes back since it did not work, I am just pushing the change for the spec for now and we can keep the other issue in the backlog to visit later.

ajkiessl (approved):
Ok sounds good.  This is good to go then!

PR #500, opened 2022-08-18
Import faculty titles form AI and display

ajkiessl:
We are now importing faculty titles from Activity Insight... again.  I removed the `ai_rank` and `ai_endowed_title` attributes for `User` that we used to import way back when.  They are no longer being used.  All we need now is `ai_title`. 

When displaying titles, `ai_title` is given preference.  If that is not present, the Organization `position_title` that is flagged as `primary` is given preference.  If that does not exist, the Organization `position_title` with the most recent `started_on` date is chosen.

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -335,7 +335,7 @@ class ActivityInsightDetailUser
   end
 
   def ai_title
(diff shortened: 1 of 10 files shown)


banukutlu (approved):
LGTM 👌🏻

PR #503, opened 2022-08-22
Fix status import for publication to grab the first status in list

ajkiessl:
Parsing the xml for status returns an array with all the statuses (including statuses for contributors).  We only want the first status.


banukutlu (approved):
looks good 👍🏻 🚢

PR #505, opened 2022-08-23
Import names from PsuIdentity

ajkiessl:
This removes the psu_identity update method from the User model and moves that logic over to PsuIdentityUserService.  Also, the original logic for this update only updated the psu_identity attribute for Users.  Now it also updates the User's first and last name with the preferred first and last name if present.  Otherwise, it updates first and last name with the regular first and last name in psu identity services.  

I regenerated a handful of vcr mocks for this.  It probably wasn't necessary, but while I was troubleshooting it just kind of happened.


ajkiessl:
Converting back to draft.  I need to make sure other importers (specifically the AI import) won't interfere with importing of names.

ajkiessl:
Ok, this is ready to go.  I changed the Activity Insight importer and Pure importer to only import user data when creating new users.  The PsuIdentityUserService now also imports middle names.

banukutlu (approved):
👌🏻  :shipit:

PR #514, opened 2022-08-26
fixing part of a11y contrast errors for links and btns

### LINKED ISSUE #55: Increase text color contrast on profile demo page for better accessibility
Some of the text/background color combinations don't provide enough contrast meet accessibility requirements.

This should be done both in the rails app and on the static profile demo page.

Aha! Link: https://penn-state-university-libraries.aha.io/features/MDDB-4

banukutlu:
Closes #55


ajkiessl (approved):
Looks good!

PR #517, opened 2022-08-29
Temporary tasks to get lost scholarsphere oals

ajkiessl:
@banukutlu I'm just going to deploy this branch temporarily to production (after I test it on qa), run the task, and then redeploy the last release once its done.  Then, I'll just delete this branch since it is only meant to be temporary.

@anaelizabethenriquez There were only 81 Scholarsphere Open Access Locations with an associated Scholarsphere Work Deposit that had no DOI.  I was only able to identify 2 of the three Scholarsphere Work Deposits with no DOI and an associated publication that has no Scholarsphere OAL.  The Scholarsphere Work Deposit IDs for these two are 1138 and 42.  Not sure what the issue is with these.  I'm also not sure why the third one isn't popping up in my queries.  I must be forgetting something 🤔 .

Edit: The third one has a Scholarsphere Work Deposit ID of 201.  It wasn't showing up in my queries because this publication doesn't have any OALs.

### DIFF
--- a/lib/tasks/database_data.rake
+++ b/lib/tasks/database_data.rake
@@ -57,7 +57,7 @@ namespace :database_data do
     filename = File.basename(Dir["#{Rails.root}/tmp/psql-rmd-prod-data-*.sql.gz"].first, '.gz')
 
     # Get local db config
-    db_config = YAML.load_file("#{Rails.root}/config/database.yml")
+    db_config = YAML.safe_load(File.open("#{Rails.root}/config/database.yml"), aliases: true)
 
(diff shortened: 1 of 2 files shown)


anaelizabethenriquez:
@ajkiessl I investigated the three ScholarSphere work deposits for which the associated publication didn't have a ScholarSphere link. The articles all turned out to have DOIs, so I added those in ScholarSphere and (if necessary) in RMD. I'll check on them after the next ScholarSphere import to make sure RMD picks up the ScholarSphere links.

PR #518, opened 2022-08-30
Unpaywall Importer: InvalidURlErrors fix

### LINKED ISSUE #341: Unpaywall Importer: Investigate InvalidURIErrors
During the Unpaywall import process, we get a bunch of `URI::InvalidURIError`s.

In the importer, we are calling `URI::DEFAULT_PARSER.escape("https://api.unpaywall.org/v2/#{publication.doi_url_path}?email=openaccess@psu.edu")` to escape the URL for the Unpaywall endpoint. In the OAB importer, we call `URI.encode`, which is essentially the same thing. The reason we call `URI::DEFAULT_PARSER.escape` is just to appease Rubocop since `URI.encode` is deprecated.

An example of a "bad" URI is https://api.unpaywall.org/v2/10.2193/0022-541X(2005)069[1211:DWDAPF]2.0.CO;2?email=openaccess@psu.edu

Notice that if you follow this link, it actually resolves and returns data from Unpaywall, so it is not actually a bad link. It may be the case that we can just remove the `URI::DEFAULT_PARSER.escape` call altogether and just hit these URLs without escaping at all.

banukutlu:
Closes #341

### DIFF
--- a/app/importers/unpaywall_publication_importer.rb
+++ b/app/importers/unpaywall_publication_importer.rb
@@ -89,7 +89,14 @@ class UnpaywallPublicationImporter
     end
 
     def query_unpaywall_for(publication)
-      find_url = URI::DEFAULT_PARSER.escape("https://api.unpaywall.org/v2/#{publication.doi_url_path}?email=openaccess@psu.edu")
+      unpaywall_url = "https://api.unpaywall.org/v2/#{publication.doi_url_path}?email=openaccess@psu.edu"
+
+      find_url = if %w[[ ]].any? { |s| publication.doi_url_path.include?(s) }
+                   Addressable::URI.encode(unpaywall_url)
+                 else
+                   URI::DEFAULT_PARSER.escape(unpaywall_url)
+                 end
+
       JSON.parse(HttpService.get(find_url))
     end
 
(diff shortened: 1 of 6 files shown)


app/importers/unpaywall_publication_importer.rb
```diff
@@ -89,7 +89,14 @@ def update_publication(publication, unpaywall_json)
     end
 
     def query_unpaywall_for(publication)
-      find_url = URI::DEFAULT_PARSER.escape("https://api.unpaywall.org/v2/#{publication.doi_url_path}?email=openaccess@psu.edu")
+      unpaywall_url = "https://api.unpaywall.org/v2/#{publication.doi_url_path}?email=openaccess@psu.edu"
+
+      find_url = if %w[[ ]].any? { |s| publication.doi_url_path.include?(s) }
+                   Addressable::URI.encode(unpaywall_url)
+                 else
+                   URI::DEFAULT_PARSER.escape(unpaywall_url)
```
banukutlu, line 97: @ajkiessl we can change this to Addressable::URI.encode, it works it that way but I wanted to get your thoughts, right now I just did it for the URL that has brackets in them.

ajkiessl: I think it's fine to just use the Addressable::URI.encode.  If you run the Unpaywall import with just the Addressable code, does it cause any errors?

ajkiessl: Also, idk if this would help with anything, but we might not need to escape the entire url.  We might just need to escape the `#{publication.doi_url_path}` part.

PR #520, opened 2022-08-31
update contribution types subject for oa policy

### LINKED ISSUE #178: Change which Contribution Types are included in the open access policy workflow
Remove "Trade Journal Article" from the OA policy workflow
Add "Conference Proceeding" to the OA policy workflow.

banukutlu:
Closes #178

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -119,7 +119,7 @@ class User < ApplicationRecord
     joins(:authorships, :user_organization_memberships, :publications)
       .joins('LEFT OUTER JOIN open_access_locations ON open_access_locations.publication_id = publications.id')
       .where(publications: { status: Publication::PUBLISHED_STATUS })
-      .where("publications.publication_type ~* 'Journal Article'")
+      .where("publications.publication_type IN (#{User.oa_publication_type_params})", *Publication.oa_publication_types)
       .where('publications.id NOT IN (SELECT publication_id from authorships WHERE authorships.id IN (SELECT authorship_id FROM internal_publication_waivers))')
       .where(%{publications.id NOT IN (SELECT publication_id from authorships WHERE authorships.id IN (SELECT authorship_id FROM scholarsphere_work_deposits WHERE status = 'Pending'))})
       .where('users.open_access_notification_sent_at IS NULL OR users.open_access_notification_sent_at < ?', 6.months.ago)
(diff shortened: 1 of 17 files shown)


ajkiessl (changes requested):
There's a couple more things scattered around that will need to be changed.

We'll need to change the logic for this method:
https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/publication.rb#L542-L544
And probably also change the name of it.  This is implemented in a couple places like the links on the open access publications index and in duplicate merging.

We also have some tasks to generate works.  The logic in these and probably the naming needs changed:
https://github.com/psu-libraries/researcher-metadata/blob/main/lib/tasks/generate_works.rake
https://github.com/psu-libraries/researcher-metadata/blob/main/lib/utilities/works_generator.rb
https://github.com/psu-libraries/researcher-metadata/blob/main/spec/factories/publication.rb#L31-L41

I like the 'oa_publication' naming, so I think we should continue to go with that instead of 'journal_article'.

ajkiessl (changes requested):
This is looking good.  There's one last thing I almost forgot about.  The query here needs to be changed to implement the new logic: https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/user.rb#L122

I believe we had set this up as one big SQL query for performance reasons.  It may be difficult to convert this line to a SQL query that can handle the new logic.  I don't think we can call the oa_publication scoping method on this since it is querying Users.  Let me know if you need any help on this.

Also (sorry last thing), there is a section in the README about Articles vs Non-article publications.  This can either be removed, or changed to fit our new logic. https://github.com/psu-libraries/researcher-metadata#journal-article-vs-non-article-publications

app/models/user.rb
```diff
@@ -119,7 +119,7 @@ def self.needs_open_access_notification
     joins(:authorships, :user_organization_memberships, :publications)
       .joins('LEFT OUTER JOIN open_access_locations ON open_access_locations.publication_id = publications.id')
       .where(publications: { status: Publication::PUBLISHED_STATUS })
-      .where("publications.publication_type ~* 'Journal Article'")
+      .where("publications.publication_type IN (#{User.oa_publication_type_params})", *Publication.oa_publication_types)
```
ajkiessl, line 122: I like what you did here ☝️ 👌

ajkiessl (approved):
Looks good, thanks!

PR #521, opened 2022-09-01
pure imports title: subtitle

### LINKED ISSUE #396: Append subtitle to title when importing Pure publications
Currently, we are importing the `subTitle` from Pure into the `secondary_title` field for `Publications`.  There seems to be a preference for this `subTitle` to be appended to the title with a colon.  So, a publication with the title: 'Math' and subtitle: 'An Interesting Subject', should be imported to the title field as 'Math: An Interesting Subject'.  The secondary title would then be left blank.

banukutlu:
Closes #396

### DIFF
--- a/app/importers/pure_publication_importer.rb
+++ b/app/importers/pure_publication_importer.rb
@@ -111,10 +111,8 @@ class PurePublicationImporter < PureImporter
   private
 
     def pub_attrs(publication)
-      subtitle = publication['subTitle'].try('[]', 'value')
-      subtitle = subtitle.present? ? ": #{subtitle}" : ''
       {
-        title: publication['title']['value'] + subtitle,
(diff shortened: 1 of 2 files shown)


app/importers/pure_publication_importer.rb
```diff
@@ -111,9 +111,11 @@ def record_type
   private
 
     def pub_attrs(publication)
+      subtitle = publication['subTitle'].try('[]', 'value')
+      subtitle = subtitle.present? ? ": #{subtitle}" : ''
```
ajkiessl, line 115: I know this is a little picky but can you refactor this into its own private method?

Otherwise, this all looks good!

banukutlu: oh yeah that makes sense, not picky at all :)

ajkiessl (approved):
Looks good!

PR #522, opened 2022-09-01
463 email conf

### LINKED ISSUE #463: Automatically send open access reminder emails weekly
- Should be timed so imports have all run just before the email goes out. 
- Emails should not go out if the most recent OA Button and/or Unpaywall imports failed or have a certain amount of errors.
- Sending these weekly is ideal for smoothing out the workload on answering support tickets. However, we may want to be able to cancel them for a particular week if there are vacations or holidays.
- Ideally the auto-deduplication (#468 ) would run before each email batch.
- It would be nice to cap the total number of emails that can be sent in one week. Otherwise, we'll send a very large batch exactly six months after the last large batch.

ajkiessl:
-Adds a singleton OaNotificationSetting class to configure the email cap for OA emails and toggle emailing on and off. 
-Implements the email cap and on/off switch
-Adds a rake task for running OA email task with the cap

closes #463

### DIFF
--- a/spec/component/models/oa_notification_setting_spec.rb
+++ b/spec/component/models/oa_notification_setting_spec.rb
@@ -49,9 +49,16 @@ RSpec.describe OaNotificationSetting, type: :model do
 
   describe 'validation' do
     context 'when creating a record when another already exists' do
-      it 'raises an error' do
+      it 'raises a RecordNotUnique error' do
         described_class.instance
-        described_class.create email_cap: 100, is_active: true
+        expect { described_class.create email_cap: 100, is_active: true, singleton_guard: 0 }
+          .to raise_error ActiveRecord::RecordNotUnique
+      end
+    end
+
+    context 'when creating a record with a singleton_guard that is not 0' do
+      it 'raises a RecordInvalid error' do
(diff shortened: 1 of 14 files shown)


spec/component/models/oa_notification_setting_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect(described_class.not_active?).to be true
+      end
+    end
+  end
+
+  describe 'validation' do
+    context 'when creating a record when another already exists' do
+      it 'raises an error' do
+        described_class.instance
+        described_class.create email_cap: 100, is_active: true
```
banukutlu, line 54: @ajkiessl We need an expect here, right?

banukutlu (changes requested):
overall looks very nice 👌🏻. I just have one comment in one of the specs.

spec/component/models/oa_notification_setting_spec.rb
ajkiessl: Oh whoops yeah

banukutlu (approved):
looks great 👌🏻  :shipit:

PR #524, opened 2022-09-02
updates webmock

whereismyjetpack:
I noticed that the version of webmock we were using was yanked, so I updated it.


Gemfile.lock
```diff
@@ -680,4 +680,4 @@ DEPENDENCIES
   webpacker
 
 BUNDLED WITH
-   2.3.8
+   2.3.7
```
ajkiessl, line 683: This probably won't affect anything, but we should try not to reduce bundle versions if we don't need to.

whereismyjetpack: Good Catch. I'll patch that up

ajkiessl (approved):

PR #528, opened 2022-09-06
482 submit a url

### LINKED ISSUE #482: Edit language on "Submit a URL" box
Current text on the "Submit a URL" box on the user-facing publication page is as follows:

> If an open access version of your publication is already available anywhere on the public web, then please submit the URL here. This should be a full and valid URL that points directly to a publicly accessible copy of your publication. 

I've learned that some users are submitting URLs that they believe aren't behind a paywall (because they access them seamlessly on campus or on the VPN) but actually are. I'd like to add text like the following, at the end of the text above:

> Not sure whether your URL is publicly accessible? Use [Open Access Button](https://openaccessbutton.org/) to check, and contact us (openaccess@psu.edu) with questions.

Smullz622:
Updated language on "submit a url" box on the edit publications page and added links to Open Access Button & contact us. fixes #482

### DIFF
--- a/app/views/open_access_publications/edit.html.erb
+++ b/app/views/open_access_publications/edit.html.erb
@@ -50,7 +50,9 @@
           <p>
             If an open access version of your publication is already available anywhere on the public web,
             then please submit the URL here. This should be a full and valid URL that points directly
-            to a publicly accessible copy of your publication.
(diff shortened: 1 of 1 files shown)


banukutlu (approved):
looks good, 👏🏻  for the first PR in this repo 🍻

ajkiessl (approved):
Looks good! 🎉

PR #538, opened 2022-09-13
465 update edited data

### LINKED ISSUE #465: Publications that have been edited by admins have outdated data
As I understand it, RMD freezes data in a publication record if the record has been edited by an admin. This includes edits made during the manual de-duplication process. I've noticed a lot of records get stuck with a status of "In Press" because of this. On its own, that may not be a big problem, but I'm wondering if there are other data integrity problems related to this "freezing."

@ajkiessl

Smullz622:
changed ai and pure importers to check imported publication status and update existing pub status accordingly. fixes #465 
Updated a lot of existing tests with :count checks since new test publications were added to the ai xml and pure json

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -230,20 +230,19 @@ class ActivityInsightImporter
       else
         # If the publication has been updated by an admin, we only want to
         # overwrite a subset of its attributes.
+        #byebug
         pub_record.update!(always_update_pub_attrs(ai_pub))
       end
     end
 
     def always_update_pub_attrs(pub)
-      {
-        activity_insight_postprint_status: pub.activity_insight_postprint_status
-        if pub.published?
-          status: pub.status
-        end
-        title: pub.title
-                #check if published, if so don't update; otherwise update
-                #instead check if import pub status is published?
-      }
+      attrs = {}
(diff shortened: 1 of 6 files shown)


ajkiessl:
We'll have to consider what system (AI or Pure) we want to give preference for titles, or we'll want to only update titles when a publication does not have multiple imports.  Otherwise, we'll end up in a situation where we'll have Pure data in the title on the weekends, and AI data for the rest of the week.  We tend to give preference to Pure, like in the `PublicationMergeOnDoiPolicy`, so we could only update the title during the AI import if there isn't a Pure import attached.  We'll also want to consider what happens when there are multiple AI imports.  They could have slightly different titles, so the titles will change as they get imported and then stick on the last one.  This will happen everyday during import.  

@anaelizabethenriquez Perhaps it would be best to not allow the title to update if there are more than one associated imports?  Or only allow the title to update during the Pure import, so that Pure would be given preference?  We would then also want to update titles for any publication that only has one associated import record from AI.  This is for issue #465 .

anaelizabethenriquez:
@ajkiessl I think it would be best to "allow the title to update during the Pure import, so that Pure would be given preference" and also "update titles for any publication that only has one associated import record from AI."

ajkiessl (approved):
Looks good!  I think we've covered every permutation here... and then some 😅 .  Thanks!

PR #540, opened 2022-09-15
Validate doi in scholarsphere upload.

### LINKED ISSUE #516: Bad formatting on DOIs in "ScholarSphere work deposits"
While looking at DOIs on ScholarSphere work deposits, I noticed 15 deposits that have DOIs but the DOIs are in different formats than we usually use in RMD. I suspect this means there's no DOI validation on the deposit form in RMD. 

Can we investigate how these "bad" DOIs are getting into the system and add validation to stop it? 

Here's a list of the DOIs:
0.1111/josh.13136
10.1002/jcad.12341
10.1007/s10802-016-0166-4
10.1016/j.amepre.2011.08.012
10.1016/j.amepre.2016.09.014
10.1016/j.jpeds.2018.07.086
10.1016/j.ypmed.2016.02.038
10.1177/0017896918820558
10.1307/mmj/20205890
10.3847/1538-4357/abd324
10.5840/epoche2020107166
DOI: 10.1891/RE-20-18
DOI: 10.1177/1525822X14526543
DOI: 10.5430/jha.v4n3p20
doi.org/10.1016/j.jacceco.2021.101407

I will go into ScholarSphere and correct these on the deposits, and I'll also check the publications in RMD and make sure those have a valid DOI. (Since there's validation on that field, I'm expecting them to be either valid or blank.)

I'm putting this in the backlog for now, @ajkiessl

ajkiessl:
closes #516

### DIFF
--- a/app/models/scholarsphere_work_deposit.rb
+++ b/app/models/scholarsphere_work_deposit.rb
@@ -104,7 +104,7 @@ class ScholarsphereWorkDeposit < ApplicationRecord
   private
 
     def doi_format_is_valid
-      if !doi.nil? && !doi.empty?
+      if doi.present?
         unless doi == DOISanitizer.new(doi).url
           errors.add(:doi, I18n.t('models.publication.validation_errors.doi_format'))
         end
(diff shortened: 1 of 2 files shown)


Smullz622 (approved):
looks good to me!

PR #541, opened 2022-09-16
530 duplicate year

### LINKED ISSUE #530: Changes to duplicate grouping logic
In the current duplicate grouping logic that requires a fuzzy match on title+subtitle and a match on publication year, change the publication year requirement to require a match +/- 2. So, e.g., 2020 would match 2018, 2019, 2020, 2021, 2022.

Smullz622:
Changed logic in duplicate public grouping to allow the year to pass as long as it's within +/-2 years
fixes #530

### DIFF
--- a/spec/component/models/duplicate_publication_group_spec.rb
+++ b/spec/component/models/duplicate_publication_group_spec.rb
@@ -526,8 +526,49 @@ describe DuplicatePublicationGroup, type: :model do
       it 'groups the publications' do
         described_class.group_duplicates_of(p1)
         group = p1.reload.duplicate_group
-
         expect(p2.reload.duplicate_group).to eq group
+      end
+    end
+
+    context 'given a publication with a similar title to another publication and a different publication date but the year is within +/-2' do
+      let!(:p1) { create :publication,
+                         title: 'A Typical Title',
+                         published_on: Date.new(2000, 5, 20),
+                         doi: nil }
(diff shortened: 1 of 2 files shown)


spec/component/models/duplicate_publication_group_spec.rb
```diff
@@ -526,11 +526,54 @@
...
+                         published_on: Date.new(2002, 1, 1),
+                         doi: nil }
+      let!(:p3) { create :publication,
+                         title: 'The Typical Title 3',
+                         published_on: Date.new(2000, 5, 20),
+                         doi: nil }
 
+      it 'groups the publications' do
+        described_class.group_duplicates_of(p1)
+        group = p1.reload.duplicate_group
```
ajkiessl, line 550: I'm tempted to have a 'not be nil' check in here, but I can go either way.  This test works assuming p1 and p3 end up grouping too, which isn't guaranteed if we are making changes to the grouping logic and something goes wrong.  It may be unlikely, but we could still end up with a nil == nil situation.

ajkiessl (dismissed):
This looks good.  I added one comment, but it's not necessary to make a change.  I'll leave that up to you.

ajkiessl:
This should be good once the conflict is resolved.

banukutlu (approved):
lgtm 👍🏻

PR #543, opened 2022-09-20
534 automerger fuzzy match

### LINKED ISSUE #534: Add title fuzzy match requirement to AI-Pure automerger
The `auto_merge` method in DuplicatePublicationGroup should query the group with the same fuzzy title matching logic as the grouping method to check if the two publications in the group match.  If so, allow the merge.  Otherwise, do not merge.

Should be done before #531

Smullz622:
Added 60% fuzzy title match to AI-Pure auto merger
fixed existing tests that were giving false positives
fixes #534


ajkiessl (approved):
LGTM!

PR #544, opened 2022-09-20
532 missing doi automerge

### LINKED ISSUE #532: Create new automerge process
In addition to the existing AI-Pure automerge process and the DOI-based automerge process, we want to create another process that matches when the publications are very similar but one or both don't have DOIs. Like the DOI-based automerger, it should review (and, if appropriate, merge) each pair of publications in a duplicate group.

The starting point for the criteria should be what's currently in the DOI-merge process (prior to the changes in #529 ), but with the following updates:

1. Title and subtitle: Add requirement that the combined title and subtitle of the two publications fuzzy match at a very high percent -- perhaps 90 or 95%.
2. DOI: If the DOIs conflict, do not merge. OK to merge if neither pub has a DOI or if only one pub does.

Smullz622:
Created new automerge to merge if one or both DOIs missing but other attributes are very similar
Added logic to dup pub grouping to determine which policy to use
Added doi logic to pub merge
fixes #532

### DIFF
--- a/app/models/duplicate_publication_group.rb
+++ b/app/models/duplicate_publication_group.rb
@@ -130,8 +130,7 @@ class DuplicatePublicationGroup < ApplicationRecord
               pub.imports.each do |i|
                 i.update!(auto_merged: true)
               end
-              pub_primary.merge_on_doi!(pub)
-              # search for all uses
+              pub_primary.merge_on_matching!(pub)
             end
           rescue Publication::NonDuplicateMerge; end
         end
--- a/app/models/duplicate_publication_group.rb
+++ b/app/models/duplicate_publication_group.rb
@@ -93,12 +93,12 @@ class DuplicatePublicationGroup < ApplicationRecord
     end
   end
 
-  def self.auto_merge_on_doi
+  def self.auto_merge_matching
     pbar = ProgressBarTTY.create(title: 'Auto-merging duplicate groups on doi',
                                  total: count)
 
     find_each do |g|
-      g.auto_merge_on_doi
+      g.auto_merge_matching
       pbar.increment
     end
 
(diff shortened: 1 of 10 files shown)


spec/component/models/duplicate_publication_group_spec.rb
```diff
@@ -1393,18 +1393,59 @@
       let!(:import1) { create :publication_import, publication: pub1 }
       let!(:import2) { create :publication_import, publication: pub2 }
 
-      context 'when PublicationMatchOnDoiPolicy returns true for these publications' do
+      context 'when both DOIs are present and equal' do
```
ajkiessl, line 1396: For the next four contexts, the tests are good but I think the context descriptions might need to be modified.  I think we need to clarify that these have been merged because they passed one of the matching policies. In the last context, we should indicate that one of the publications did not pass either match policy so it did not merge.

app/models/duplicate_publication_group.rb
```diff
@@ -119,14 +119,19 @@ def auto_merge_on_doi
...
+                         PublicationMatchMissingDoiPolicy.new(pub_primary, pub)
+                       end
         if match_policy.ok_to_merge?
           begin
             ActiveRecord::Base.transaction do
               pub.imports.each do |i|
                 i.update!(auto_merged: true)
               end
               pub_primary.merge_on_doi!(pub)
+              # search for all uses
```
ajkiessl, line 134: Can this comment be removed?

ajkiessl (dismissed):
Looks good.  I just have a few comments about changes.

banukutlu (approved):
looks good 👍🏻  just one very minor refactor for journal title check that we talked with Sarah.

PR #546, opened 2022-09-21
531 doi grouping

### LINKED ISSUE #531: Add new duplicate grouping logic
In addition to the current duplicate grouping logic (updates to which are described in #530), add another way duplicates are grouped, based on DOI.

If a publication has the same DOI as another publication, put them in a duplicate group together. If either or both of the publications are already in duplicate groups, combine the duplicate groups.

Note that we may need to adjust the timing of when the current AI-Pure automerger runs to ensure that this additional grouping doesn't interfere with that automerge process (which will only merge if the duplicate group contains two pubs with one import each). That process probably needs to run after each AI import and before this new grouping process. When we roll out these changes, we should run the revised duplicate grouping process (from #530 ), then run the AI-Pure automerger, then run this new process.

Blocked by #529 and #534

Smullz622:
added logic to group_duplicates_of to add a publication to duplicates if only doi matches
fixes #531

### DIFF
--- a/spec/component/models/duplicate_publication_group_spec.rb
+++ b/spec/component/models/duplicate_publication_group_spec.rb
@@ -585,7 +585,7 @@ describe DuplicatePublicationGroup, type: :model do
                          published_on: Date.new(2003, 1, 1),
                          doi: 'https://doi.org/10.000/some-doi-22357534' }
 
-      it 'groups the publications' do
+      it 'does not group the publications' do
         described_class.group_duplicates_of(p1)
         group = p1.reload.duplicate_group
 
--- a/spec/component/models/duplicate_publication_group_spec.rb
+++ b/spec/component/models/duplicate_publication_group_spec.rb
@@ -246,7 +246,7 @@ describe DuplicatePublicationGroup, type: :model do
       let!(:p15_1) { create :publication,
                             title: 'A Perfect Title Match Where a DOI is blank',
                             published_on: Date.new(2000, 1, 1),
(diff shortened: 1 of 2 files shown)


spec/component/models/duplicate_publication_group_spec.rb
```diff
@@ -554,6 +554,46 @@
...
+                         title: 'A Generic Title',
+                         published_on: Date.new(2000, 1, 1),
+                         doi: 'https://doi.org/10.000/some-doi-22357535' }
+
+      let!(:p2) { create :publication,
+                         title: 'A Different One',
+                         published_on: Date.new(2003, 1, 1),
+                         doi: 'https://doi.org/10.000/some-doi-22357534' }
+
+      it 'groups the publications' do
```
ajkiessl, line 588: I think we want 'does not group the publications' here.

ajkiessl (commented):
This looks good.  There's a comment I made in the issue #531 about implementing the grouping and autodeduping methods in the importers.  We can hold off on that for now.  It would probably be better to get what we have out and running to see how well it works before we get so fine-tuned with the import autodeduplication.  I'll make a separate ticket for that.  If we end up doing that, though, we'll have to separate the doi grouping from the title/pub year grouping.

spec/component/models/duplicate_publication_group_spec.rb
Smullz622: yep, my bad. I just updated it

ajkiessl (approved):
lgtm!

PR #547, opened 2022-09-22
Hot fixes

ajkiessl:
When the deposit record was being use, the form would disable when a validation error occurred.  Now using authorship record instead of deposit record to determine if doi is present.

Also fixed up some tests for the WorksGenerator.


banukutlu (approved):
👍🏻

PR #553, opened 2022-09-27
162 oab on title

### LINKED ISSUE #162: Look up Publications in OA Button and/or Unpaywall based on fields other than DOI
Two reasons to do this:

1. RMD doesn’t have the DOI (maybe it doesn’t exist) 
2. The repository doesn’t have the DOI (but RMD might) 

URL, DOI, and Title all appear to be options for querying the OA Button API. Not sure about Unpaywall.

Smullz622:
changed logic in oabutton importer and unpaywall importer to allow a search by title when doi is missing
since unpaywall can return several results, it has additional title validation logic
fixes #162


ajkiessl:
I haven't reviewed all of this yet but was just thinking something.  If Unpaywall or OAB are providing DOIs from these requests, we should probably grab them if we don't have them in RMD.  This doesn't have to be a part of this PR, though, I will make a separate ticket for this.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -110,4 +131,9 @@ def update_open_access_location(open_access_location, unpaywall_json)
     def build_new_oal(publication, url)
       publication.open_access_locations.build(source: Source::UNPAYWALL, url: url)
     end
+
+    def title_match?(title1, title2)
+      title1&.downcase&.gsub(/[^a-z0-9]/, '')&.include?(title2&.downcase&.gsub(/[^a-z0-9]/, '')) &&
+        title2&.downcase&.gsub(/[^a-z0-9]/, '')&.include?(title1&.downcase&.gsub(/[^a-z0-9]/, ''))
```
ajkiessl, line 137: Do we want to use `||` instead of `&&` here?  The way it would work right now I think would basically require and exact match.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -81,16 +91,27 @@ def update_publication(publication, unpaywall_json)
         locations_to_delete = existing_locations.reject { |l| unpaywall_locations_by_url.key? l.url }
         locations_to_delete.each(&:destroy)
 
-        publication.open_access_status = unpaywall_json['oa_status']
+        publication.open_access_status = if publication.doi.present?
+                                           unpaywall_json['oa_status']
+                                         else
+                                           unpaywall_json['results'].first['response']['oa_status']
```
ajkiessl, line 97: If we send a title search to unpaywall and it returns nothing, do we still receive the "results" key in the hash?  If it does return the "results" key, but there's nothing in it, we may run into some NoMethod for nil errors when trying to grab the 'response' and 'oa_status'.  If you run the imports locally, do you see a lot of these errors?

spec/fixtures/vcr_cassettes/UnpaywallPublicationImporter/_import_all/when_an_existing_publication_does_not_have_a_DOI/does_not_create_any_open_access_locations_for_the_publication.yml
```diff
@@ -0,0 +1,213 @@
+---
```
ajkiessl, line 1: I think some of these vcr cassettes can be deleted, like this one.  They might have been created when the tests were ran after the importers were changed but the tests weren't changed.  Looks like there are about 28 unpaywall tests that send requests, so there should be about the same number of vcr cassettes.

ajkiessl (changes requested):
I have a couple questions/change requests for this one.

app/importers/unpaywall_publication_importer.rb
Smullz622: I did that to account for short titles that show up on some publications. Ex: 'Economic Development' matched for 'Economic Development, A Reflection', 'The Role of Agriculture in Economic Development', etc. Ideally we would fuzzy match, but from what I saw, it looked like we would need other gems to do so outside of a Postgres query

ajkiessl: So if we are checking that title 1 is included in title 2 _and_ title 2 is included in title 1, aren't we just doing an `==` operation?

app/importers/unpaywall_publication_importer.rb
Smullz622: I didn't run the imports locally because it seemed like it would take a few days. Since it matches on keywords, I don't think it would occur that often, but I can a nil check before assigning unpaywall_title just to be safe

app/importers/unpaywall_publication_importer.rb
Smullz622: Yep. My bad. I can clean that up

ajkiessl (approved):
Looks good!  I removed some of the vcr cassettes that weren't being used.

PR #556, opened 2022-09-28
542 exclude inactive users

### LINKED ISSUE #542: Exclude inactive users from #needs_open_access_notification query
Right now we do not exclude inactive (as defined by their PSU Identity status) users from this query.  We simply do not send them an email when the emails go out.  We are starting to accumulate inactive users.  If this number gets too large, they could clog up our emailing system.  Excluding them in the query will fix this.

Smullz622:
added logic to exclude inactive users from needs_open_access_notification query
fixes #542


spec/component/models/user_spec.rb
```diff
@@ -421,7 +421,10 @@
 
   describe '.needs_open_access_notification' do
     # Users who meet all of the criteria to receive an email
-    let!(:email_user_1) { create :user, open_access_notification_sent_at: 1.year.ago, first_name: 'email_user_1' }
+    let!(:email_user_1) { create :user,
+                                 open_access_notification_sent_at: 1.year.ago,
+                                 first_name: 'email_user_1',
+                                 psu_identity: { 'data' => { 'link' => '', 'cprid' => '', 'active' => true, 'userid' => '', 'confHold' => false, 'givenName' => '', 'familyName' => '', 'affiliation' => ['FACULTY', 'MEMBER'], 'displayName' => '', 'serviceAccount' => false, 'universityEmail' => '' } } }
```
ajkiessl, line 427:
```suggestion
 psu_identity: { 'data' => { 'link' => '', 
                              'cprid' => '', 
                              'active' => true, 
                              'userid' => '', 
                              'confHold' => false, 
                              'givenName' => '', 
                              'familyName' => '', 
                              'affiliation' => ['FACULTY', 'MEMBER'], 
                              'displayName' => '', 
                              'serviceAccount' => false, 
                              'universityEmail' => '' }
                     } 
               }
```
For hashes, it's sometimes helpful to add a new line for each key/value and align them so they don't run off the screen.

ajkiessl (changes requested):
This looks good.  The only thing I would suggest is to add new lines for each key/value to the psu identity hashes in the tests.  This will make them more readable.

ajkiessl (approved):

PR #559, opened 2022-09-29
blocked email notification if ai post print status is 'in progress'

### LINKED ISSUE #113: Automate data that prevents nudge email
In addition to articles with OA link in RMD, the following articles should not get a nudge:

- Article is deposited in ScholarSphere with or without embargo
- Article is in process of deposit to ScholarSphere via AI
- There is a waiver for article (through RMD or alternate form)

For the initial batch, we'll do this manually, but it would be nice to automate what we can, down the line.

Smullz622:
Articles in process of deposit to ScholarSphere via AI should not get a nudge. If this is the case, activity_insight_postprint_status will be 'In Progress' and nudge is blocked.
fixes #113

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,6 +38,10 @@ RSpec/LeakyConstantDeclaration:
   Exclude:
     - 'spec/component/models/concerns/null_object_pattern_spec.rb'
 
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
Looks good, thanks!

PR #561, opened 2022-09-29
updated doi grouping logic so empty string DOIs won't match

### LINKED ISSUE #558: `doi = ''` grouping logic
Our new grouping logic will group publications that have the same dois.  It won't group publications whose dois equate to `nil == nil` (which is good), but it does group publications whose dois equate to `'' == ''`.  

We need to add to the SQL query to not group when dois are empty strings ('').  We every so often pull in data like this, so we have about 600 that get added to a duplicate group when the new grouping logic is run.

Smullz622:
fixes #558


ajkiessl (approved):
LGTM Thanks!

PR #565, opened 2022-10-03
85 publications validations

### LINKED ISSUE #85: Publications Validations
Validations for ISBN/ISSN/DOIs.  This can either be on the model level or at import.  

Having model level validations will give us more universal control over what gets into the database.  As of right now our biggest issue is data coming in from Digital Measures.  Sometimes the data is either wrong, or placed in the wrong field (ie a doi in the isbn/issn field).  This is where validating on import might be a better solution, since we could detect identifiers in the wrong field and put them in the right place on import.

It might make sense to implement both of these.  Model level validations for universal use, and something that detects identifiers in the wrong fields during the DM import.

Smullz622:
added ISSNSanitizer and ISBNSanitizer classes to check for and validate issn & isbn in ISBNISSN, DOI, and url fields in Activity Insight import
fixes #85

### DIFF
--- a/app/models/issn_sanitizer.rb
+++ b/app/models/issn_sanitizer.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ISSNSanitizer
+    def initialize(value)
+      @value = value
+    end
+  
+    def issn
+      return nil if ISBNSanitizer.new(value).isbn || DOISanitizer.new(value).url
+      issn_match = /(ISSN |eISSN )?[\S]{4}\-[\S]{4}/.match(value)
+    end
+  
+    private
+  
+      attr_reader :value
+  end
+  
--- a/app/models/issn_sanitizer.rb
+++ b/app/models/issn_sanitizer.rb
@@ -1,17 +1,17 @@
 # frozen_string_literal: true
 
 class ISSNSanitizer
-    def initialize(value)
-      @value = value
-    end
-  
-    def issn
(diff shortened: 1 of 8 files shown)


app/models/issn_sanitizer.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class ISSNSanitizer
+  def initialize(value)
+    @value = value
+  end
+
+  def issn
+    return nil if ISBNSanitizer.new(value).isbn || DOISanitizer.new(value).url
+
+    /(ISSN |eISSN )?\S{4}-\S{4}/.match(value)
```
ajkiessl, line 11: I ran the AI import and found this regex was pulling things like this from urls: 'word-word'.

We should probably make this more restrictive with a regex like this: `(ISSN |eISSN )?\d{4}-\d{3}(\d|x|X)`

ajkiessl (changes requested):
I added a change request about the ISSN regex.  I think it might need to be a bit more restrictive.  

Also, I know we are kind of testing the functionality of the sanitizers in the AI importer tests, but I think we should add some unit tests for the sanitizers specifically (like how they are done for the DoiSanitizer: https://github.com/psu-libraries/researcher-metadata/blob/main/spec/unit/models/doi_sanitizer_spec.rb).  We could then test some of the extra logic in the ISSN and ISBN sanitizers like when it detects an ISBN in the ISSNSanitizer and returns nil.  Let me know if you have any questions.

PR #566, opened 2022-10-03
Assess article version during SS deposit

### LINKED ISSUE #467: Assess article version during deposit process
When a user uploads a file to deposit, attempt to check the article version. If one can be determined, display it to the user and give deposit advice accordingly (this could be very general or could be based on #111 ). If version cannot be determined, tell the user that and perhaps allow them to indicate an article version for the purposes of #111. If user is trying to upload something other than an accepted manuscript or a final published version, encourage them not to. (OA policy requires accepted manuscript. Final published version is even better, so we'll take that if they are allowed to share it.)

banukutlu:
Closes #467


banukutlu:
This work is merged into #569, it will be reviewed and closed with that PR.

PR #568, opened 2022-10-04
536 pure 404 email

### LINKED ISSUE #536: Email when Pure importer service not found
When we get a 404 from Pure, the error is being caught.  It gets logged to our import error logs, but its not so easily noticed.  We may want to send an email to the dev team when this happens, or find some other way to notify (maybe datadog?)

Smullz622:
added #pure_import_error to admin_notifications_mailer
added new logic to pure_importer to call the mailer
fixes #536


ajkiessl (approved):
LGTM thanks!

PR #570, opened 2022-10-05
555 grab dois

### LINKED ISSUE #555: Grab DOIs from Unpaywall and Open Access Button
For records in RMD that don't have DOIs, we send requests to Unpaywall and OAB to try to find OA Urls.  If Unpaywall or OAB is also providing a DOI, we should grab it and store it on import.

Smullz622:
When publications without DOIs title match during oa location searches in OAButton or Unpaywall , the doi is retrieved and saved in the publication.
fixes #555


app/importers/open_access_button_publication_importer.rb
```diff
@@ -70,4 +73,8 @@ def query_open_access_button_for(publication)
         }
       )
     end
+
+    def doi
+      DOISanitizer.new(doi_element).url || DOISanitizer.new(url).url || DOISanitizer.new(issn).url
+    end
```
ajkiessl, line 79: I think this can be removed since it isn't doing anything.

ajkiessl (changes requested):
Looks good!  I just have one small change request.

app/importers/open_access_button_publication_importer.rb
Smullz622: Yep, my bad, I meant to delete that

ajkiessl (approved):

PR #574, opened 2022-10-10
Unpaywall import changes

### LINKED ISSUE #572: Nil handling Unpaywall import
undefined method `[]' for nil:NilClass unpaywall_title = unpaywall_json['results'].nil? ? '' : unpaywall_json['results'].first['response']['title']

I don't think this is preventing anything from being imported, but it does add a lot of errors to the error log.  Would be best to resolve it.

ajkiessl:
I made some changes here to help us out with the "new" unpaywall imports.  I added nil handling for the results we get from title searches.  An error was being raised for a lot of these, which then prevents the unpaywall_last_checked_at timestamp from being set to true.  This could lead to a buildup of these whenever we run the "new" imports, so I added the nil handling to prevent that.

Also, we should be updating the unpaywall_last_checked_at timestamp for pretty much every publication that we send a request to unpaywall, now (unless there's an error).  Even if we don't get anything from unpaywall, we still want to indicate that it was checked.

I also added a CGI.escape around the title in the url for the request to unpaywall.  Since we were getting a good amount of invalid url errors.

closes #572


banukutlu (approved):
👌🏻

Smullz622 (approved):

PR #583, opened 2022-10-24
Only Group Duplicates when new publication during AI import

### LINKED ISSUE #577: New Duplicate Grouping Logic and Activity Insight Imports
After keeping an eye on the AI import with the new grouping logic, I've noticed a few things and would like to make some changes in the near term.  This isn't necessarily urgent (everything is fine the way it is right now), however, I think I'd like to address this ASAP after I return from vacation.

The average latency of the new grouping logic query is about 2.8 seconds.  That adds up during the AI import since we are importing several hundred thousand publications.  The grouping logic happens on every imported publication, regardless of whether it is a new publication or an existing publication.  This is what I'm thinking of changing.

After running the Pure import, I noticed it didn't take much longer than it did before we added the new grouping logic.  Looking through the code again, I realized we _don't_ run the grouping logic on every publication import from Pure, only the new publications.  I think this is a better way to handle grouping.  We don't need to run the grouping logic on _every_ publication unless we've changed the grouping logic.  So, we could really improve the performance of the AI import by doing it like the Pure importer.  We could then return to running the AI import more frequently like we were doing before the last release.

Moving forward I think we should make sure all our importers only do the grouping logic on new publications.  This will require updating the PSU Law importer and the Web of Science importer.  Every time we update our grouping logic we should run the rake task that iterates over all our stored publications and runs the grouping logic on them.  

When I return from vacation, we'll want to run the rake task that iterates over all publications so the existing Pure, PSU Law, and Web of Science publications get grouped with the new grouping logic.  The AI import will run on Monday (Oct 17th) with the grouping logic still acting on every publication.  This is fine, it will just take a long time.  It looks like it will take about a week to finish.

@anaelizabethenriquez Tagging you here so you are aware.

ajkiessl:
Tweaks to ai importer to only group duplicates when a new publication is created and tweaked the duplicate grouping task to toggle publications to not be visible if they have not already been grouped.

closes #577

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -167,6 +167,11 @@ class ActivityInsightImporter
             update_pub_record(pub_record, pub)
           else
             pi.save!
+
+            DuplicatePublicationGroup.group_duplicates_of(pub_record)
+            if pub_record.reload.duplicate_group
+              pub_record.update!(visible: false)
+            end
           end
 
           if pub_record.updated_by_user_at.blank?
@@ -194,12 +199,6 @@ class ActivityInsightImporter
               c.user = u if cont.for_imported_user?
               c.save!
             end
-          end
-
-          DuplicatePublicationGroup.group_duplicates_of(pub_record)
-
-          if pub_record.reload.duplicate_group
-            pub_record.update!(visible: false)
           end
         end
       rescue StandardError => e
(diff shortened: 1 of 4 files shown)


Smullz622 (approved):
lgtm, thanks!

banukutlu (approved):
👌🏻

PR #587, opened 2022-10-25
Clean " and ' characters from OAB title query

### LINKED ISSUE #586: Clean title search to OAB
It looks like OAB is detecting signs of bots when it receives a query to its title search endpoint and blocking anything it thinks is a bot.  It seems that HTML in our titles are triggering this, but potentially also special characters.  They're block only seems to last for an hour, so luckily it is not permanent.  However, we should get this fixed ASAP so we don't get blocked permanently and so our importers aren't slowed down.  We should try stripping certain special characters (especially '<' and '>') from titles.

ajkiessl:
Closes #586


ajkiessl:
Waiting to merge this until the OAB importer completes to find any other characters that may be causing issues.  The import will still run even though we get blocked during some requests.  Seems like in some cases they throttle us after they detect us as a bot, but the import eventually picks back up.

PR #589, opened 2022-10-31
Handling integers and arrays in exif tool

ajkiessl:
Sometimes the subject fields from the exif tool can be an array, and the creator and creator_tool fields can have integers.  I modified some of the code to work with these types.  I also added some tests.


banukutlu (approved):
looks good 👌🏻 thanks!

PR #639, opened 2023-01-10
Adds ActivityInsightOaFile model and db migration.  Tests as well.

### LINKED ISSUE #638: `ActivityInsightFile` model
Create model to store post print file metadata.  This could be something like `ActivityInsightFile`.  It will belong to a `Publication` and a `Publication` can have many `PostPrintFile`s.  We'll need `location` as one of it's attributes.  

Part of #625

ajkiessl:
Closes #638


ajkiessl:
Need to add a checksum field

db/migrate/20230110174055_create_activity_insight_oa_files.rb
```diff
@@ -0,0 +1,16 @@
...
+
+      t.timestamps
+    end
+
+    add_foreign_key :activity_insight_oa_files, 
+                    :publications, 
+                    name: :activity_insight_oa_files_publication_id_fk,
+                    on_delete: :cascade
+  end
+end
```
EricDurante, line 16: @ajkiessl I think we should add an index on the foreign key column (`publication_id`) too.

spec/component/models/activity_insight_oa_file_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  subject { described_class.new }
+
+  it { is_expected.to have_db_column(:location).of_type(:string) }
+  it { is_expected.to have_db_column(:checksum).of_type(:string) }
+  it { is_expected.to have_db_column(:created_at).of_type(:datetime).with_options(null: false) }
+  it { is_expected.to have_db_column(:updated_at).of_type(:datetime).with_options(null: false) }
+
+  it { is_expected.to have_db_foreign_key(:publication_id) }
+  it { is_expected.to belong_to(:publication).inverse_of(:activity_insight_oa_files) }
+end
```
EricDurante, line 15: Since we have a cascading delete on our relationship with publication, I'd like to specify that here. Normally I'd test this by creating an `ActivityInsightOaFile` record and associated `Publication`, deleting the `Publication`, and asserting that the `ActivityInsightOaFile` has been deleted also.

EricDurante (approved):
Looks good!

PR #642, opened 2023-01-11
Adds Config gem

### LINKED ISSUE #506: 12-factor configuration
To better support our Kubernetes deployments, We should promote environment specific into ENV variables. 

we should be able to do this by copying all the `.sample` configuration files into their `.yml` counterparts, and replacing the values with ERB ENV calls. 

We could also refactor how the app pulls its configuration and use something like `figaro` or `config` This option I think will require more effort.

whereismyjetpack:
Fixes #506 

- Installs and configures the `config` gem
- moves all defaulty app configuration to `./config/settings.yml`
- moves all environment default configuration to `./config/settings/#{environment}.yml`
- removes .yml.ci configurations, as they are now in config/settings/test.yml

### Setup 
Copy default configuration, and add your local overrides.
```
cp ./config/settings.yml config/settings.local.yml
```

`test` env does not read from `./config/settings.local.yml` any test specific configuration should be added to `/config/settings/test.yml`

### Cleanup 
Once you've verified that your app is working with the new configuration, you can remove the old settings files 
```
rm  config/scholarsphere-client.yml
rm config/activity_insight.yml
rm config/azure_ad_oauth.yml
rm config/orcid.yml
rm config/pure.yml
```

### Accessing and adding Settings 
All settings are accessed by the global `Settings` object using object member notation 
e.g the following config 
```
crank:
  it:
    up: true
```
Can be accessed ala `Settings.crank.it.up`

Settings can also be added via environment variable, using `__` as the config separator, prefixed with `SETTINGS`  e.g the example above can be set like this 
```
export SETTINGS__crank__it__up=true
```

Lastly, apologies for the large PR. I tried to split it up into logical commits, and the changes are mostly renaming variables and moving files around :( 

### TODO 

- [x] update the `researcher-metadata-chef` repo with a `settings.local.yml` that cap can place
- [x] update the wiki with new instructions


EricDurante:
Looks good to me. I think that all of the separate config files for the secrets were nothing more than an artifact of our convention for organizing encrypted data bags in Chef. I've never used the Config gem, but I've used Figaro ages ago, and we got away from it in favor of the pattern that we had here, but only because of the way that we were managing server configuration. There's nothing special about it otherwise.

EricDurante:
btw, I'd approve this, but I don't seem to be able to for some reason.

whereismyjetpack:
Thanks. I have permission to approve, but not permission to give you permission to approve. @ajkiessl ?

ajkiessl:
@whereismyjetpack I upgraded the DevOps team to admin permissions.  Did that change anything?

EricDurante (approved):

EricDurante:
Great success!

PR #647, opened 2023-01-12
Mock PsuIdentityUserService in tests

ajkiessl:
Had to generate VCRs for the user_profile tests.  Also, had to mock User#from_omniauth during the stubbed authentication to stop live calls during the integration tests.

### DIFF
--- a/spec/component/models/user_profile_spec.rb
+++ b/spec/component/models/user_profile_spec.rb
@@ -2,7 +2,7 @@
 
 require 'component/component_spec_helper'
 
-describe UserProfile do
+describe UserProfile, vcr: true do
   subject(:profile) { described_class.new(user) }
 
   let!(:user) { create(:user,
--- a/spec/fixtures/vcr_cassettes/UserProfile/1_1.yml
+++ b/spec/fixtures/vcr_cassettes/UserProfile/1_1.yml
@@ -1,0 +1,55 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://identity.apps.psu.edu/search-service/resources/people/userid/abc123
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - Faraday v1.8.0
+      X-Datadog-Trace-Id:
+      - '2849599822675619865'
+      X-Datadog-Parent-Id:
+      - '197981042850209954'
+      X-Datadog-Sampling-Priority:
+      - '1'
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      Accept:
+      - "*/*"
+  response:
+    status:
+      code: 404
+      message: Not Found
+    headers:
+      Date:
+      - Thu, 12 Jan 2023 19:14:42 GMT
+      Content-Type:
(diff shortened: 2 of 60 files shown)


ajkiessl:
Converting back to draft.  There's _a lot_ more external calls that need to be mocked.

spec/component/models/user_profile_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'component/component_spec_helper'
 
-describe UserProfile do
+describe UserProfile, vcr: true do
```
EricDurante, line 5: Instead of generating all of these fixtures, what does everybody think about just mocking `PsuIdentityUserService` for this whole spec? Then if we want to test the integration from `UserProfile` all the way up to the external API call to the PSU identity service, we can write just one test to do that specifically and stick it in spec/integration. Or... if we can trust that the psu_identity gem is doing its job right, then maybe our integration test doesn't need to mock the HTTP requests at all. It can just mock `PsuIdentity::SearchService::Client`.

Even if we do generally trust the `PsuIdentity::SearchService::Client` API wrapper, I think there's definitely value in having a single integration test somewhere that exercises the interaction with the real external service. That's generally how I approach testing against external APIs when a 3rd-party  wrapper is available and we've hidden the wrapper behind our own interface (I know that you're the "3rd party" in this case, but the point is that someone has already abstracted the HTTP calls):

1. mock our own interface (`PsuIdentityUserService` in this case) for all of the unit tests
2. maybe write a few component tests that exercise the interaction between our code and the 3rd-party wrapper while mocking the wrapper (`PsuIdentity::SearchService::Client` in this case), and of course test our own interface also while mocking the wrapper
3. then, if it's feasible to test against the real external service (or an identical sandbox), write one heavy-weight integration test that exercises the whole stack IRL

I wouldn't normally write any tests that use fixtures to mock HTTP requests and responses unless I wrote the code that was making the HTTP requests.

We probably wouldn't want a full-stack integration test to run as part of our default test suite in development, but we could tag it so that it only runs on CI, and accept the tradeoff of knowing that it's going to fail occasionally due of an outage for the assurance that our app actually still works with the real PSU identity service.

whereismyjetpack: I tend to agree with this, If we keep a single integration test that confirms the shape of the response from the identity api, we can mock out the rest. I tend to get wheezy with vcr, but that can also be chalked up to my limited understanding of vcr, and not a real problem.

EricDurante: I can see a use for a tool like VCR (though I've never used it myself), but that use is really limited. I looks to me like it's easy to overuse it an a way that just ends up enabling anti-patterns.

Anyway, if it would be helpful (and if I get a chance), I could submit a separate PR to demonstrate what I'm suggesting here.

ajkiessl: Yeah I agree.  I can take a stab at this today.  `UserProfile` basically uses `PsuIdentity::SearchService::Client` every time it's instantiated, so it's sporadically sending requests throughout the test suite.

ajkiessl:
Replaced with #659

PR #652, opened 2023-01-17
makes upload queue configurable

### LINKED ISSUE #651: make queue name configurable
delayed job queues up jobs as `scholarsphere-uploads-#{`hostname`}".strip` we should make this configurable, since in kubernetes the host that puts the job is not the one that will do the job.

whereismyjetpack:
Fixes #651 

We could have probably gotten away with hard-coding this, but if we ran into a situation where we were running on multiple linux nodes we might want to make each one unique.


ajkiessl (approved):

EricDurante (approved):
Looks good to me.

PR #656, opened 2023-01-18
make action_mailer configurable

### LINKED ISSUE #655: Remove hardcoded default_url_options
We have the default_url_options hardcoded for the 3 rmd environments, I'm wondering if these should be config settings that are defaulted, and then overridden so that preview environemnts can work. 

We also probably want a way to send mail through a relay other than smtp.psu.edu for testing purposes. we use mailhog in some other environments, but that project also seems to be stagnent, we can use mailcatcher perhaps?

whereismyjetpack:
Fixes #655 

I could be convinced to promote all of the mail settings to `application.rb` and have the apps local configuration drive weather or not it's turned on. most of our other apps when deployed are 'production' even if they are 'preview' or 'qa' so, we might want to re-think how these are defaulted. this PR emulates the current behavior, and we can override in helm flipping the bits on and off.

### DIFF
--- a/config/environments/beta.rb
+++ b/config/environments/beta.rb
@@ -68,11 +68,11 @@ Rails.application.configure do
   # config.active_job.queue_name_prefix = "researcher_metadata_beta"
 
   # Mail configuration
-  config.action_mailer.perform_deliveries = true
+  config.action_mailer.perform_deliveries = Settings.action_mailer.perform_deliveries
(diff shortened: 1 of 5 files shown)


whereismyjetpack:
I moved host to a top level thing, i think it makes sense. _shrug_ naming things was the hardest part of this PR

EricDurante (approved):

PR #659, opened 2023-01-19
Use test doubles in place of PSU identity service API wrapper

EricDurante:
So this currently does the mocking part of what I was suggesting in https://github.com/psu-libraries/researcher-metadata/pull/647. I noticed that the tests for our wrapper for the psu_identity gem are using VCR cassettes to mock the HTTP responses already. So now I don't think that we have any tests that are checking our interaction with the real identity service API, and with that being the case, I still think we should write one that does. I'll try to do that if I get a chance.

### DIFF
--- a/spec/component/forms/new_deputy_assignment_form_spec.rb
+++ b/spec/component/forms/new_deputy_assignment_form_spec.rb
@@ -170,6 +170,11 @@ describe NewDeputyAssignmentForm, type: :model do
     context 'when a User exists for the given webaccess id' do
       let!(:existing_user) { create(:user, webaccess_id: deputy_webaccess_id, first_name: 'Deputy', last_name: 'FromDB') }
 
+      before do
+        person = instance_spy(PsuIdentity::SearchService::Person)
+        allow_any_instance_of(PsuIdentity::SearchService::Client).to receive(:userid).with(deputy_webaccess_id).and_return(person) # rubocop:todo RSpec/AnyInstance
+      end
+
       context 'when everything goes as expected' do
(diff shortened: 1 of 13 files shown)


ajkiessl (commented):
This all LGTM outside of the rubocop errors.

EricDurante:
> This all LGTM outside of the rubocop errors.

Yeah, rubocop thinks that some of Rspec's features are naughty and that I shouldn't use them ever under any circumstances. I respectfully disagree 😆. It's fine advice for the unit tests (in fact, maybe I'll make that one change to user_spec.rb), but doing the same thing with instance doubles wasn't going to work nearly as cleanly as this did for the integration tests (I tried).

Generally speaking, I've seen other patterns for accomplishing what I did here. They have their merits and tradeoffs, but I think I prefer this approach.

EricDurante:
@ajkiessl Oh, oops. I thought that you were referring to the new rubocop todo comments regarding RSpec/AnyInstance, so that's what I was talking about above. I missed the fact that there were new rubocop offenses after my last commit. Those are fixed now.

ajkiessl:
Ah that's my bad, I should've mentioned CI was failing.  I agree with you about the RSpec/AnyInstance, stuff though.  Sometimes `allow_any_instance_of` is just the best thing to use 🤷‍♂️ .

EricDurante:
How come CI is failing now?

whereismyjetpack:
@EricDurante looks like a permissions problem, i wonder if I kick it off if it will work

ajkiessl (approved):

PR #660, opened 2023-01-19
updates known vulns

whereismyjetpack:
I've removed the dependabot configuration until we can find a way to get it to vendor, or get something else that vendors like we do. 

I manually ran `bundle audit` for this PR. I could see us having a bundle-audit actions or CI stage that issues PRs for us If this is something we want to look into, I can do this


ajkiessl:
> I could see us having a bundle-audit actions or CI stage that issues PRs for us If this is something we want to look into, I can do this

Sure, but won't dependabot still give us PRs for known vulns (just without the vendoring)?  Or does bundle-audit do a better job of it?

whereismyjetpack:
dependabot can give prs without vendoring, we'd have to manually pull down and vendor, my gut tells me that if we ran bundl-audit in a CI stage, we could add the cached gems without needing to manually do it

ajkiessl:
Ah okay, then I think we should try the bundle-audit way.

PR #663, opened 2023-01-19
ActivityInsightOaFile Importer

### LINKED ISSUE #635: Import post print file metadata from Activity Insight
During the Activity Insight Import, if a post print file exists, create an `ActivityInsightFile` record, link it to the imported publication, and store its location and checksum.

We should be checking for a file for every publication, even if it has been edited by and admin.  We don't need to check for files for anything that is already open-access.

When a file is detected, pull the file temporarily and generate a checksum from it.  This should be stored with the file record and used to detect when we are dealing with a file that is already imported.  Of course, if the checksum of a file we detect matches the checksum of a file record already in our database, do not import it.  The process of pulling the file down, generating a checksum, checking, and storing the metadata should run in it's own job.

Part of #625 

Depends on #638

ajkiessl:
The importer will import an `ActivityInsightOaFile` if there is a 'POST_FILE' with a 'DOC' (location) for a publication in Activity Insight.  It will not import a file if a publication is already open access in RMD.  This PR also removes the checksum field for `ActivityInsightOaFile` since it isn't needed.

I added the tests for this importer to the end of spec/component/importers/activity_insight_importer_spec.rb.  This file is getting so large, it's hard to know where to put things.  It may be due for a refactoring (along with the importer).  I think it covers everything it needs to, though.

closes #635

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -201,7 +201,8 @@ class ActivityInsightImporter
             end
           end
 
-          activity_insight_file_location = pub.postprints.first.location
+          activity_insight_file_location = pub.postprints&.first&.location
+
           if pub_record.no_open_access_information? && activity_insight_file_location.present?
             aif = pub_record.activity_insight_oa_files.first
 
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -200,6 +200,20 @@ class ActivityInsightImporter
               c.save!
             end
           end
+
+          activity_insight_file_location = pub.postprints.first.location
+          if pub_record.no_open_access_information? && activity_insight_file_location.present?
+            aif = pub_record.activity_insight_oa_files.first
+
+            if aif.present?
+              aif.update location: activity_insight_file_location
+              aif.save!
+            else
+              file = ActivityInsightOaFile.create(location: activity_insight_file_location)
+              pub_record.activity_insight_oa_files << file
+              pub_record.save!
+            end
+          end
         end
       rescue StandardError => e
         log_error(pub, e, u)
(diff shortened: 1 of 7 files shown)


EricDurante:
I totally agree that this stuff is very much due for refactoring, though I'm not even quite sure how I'd approach that. I like having this amount of end-to-end test coverage for the import processes since their behavior is complicated, and it's really important to verify that it's absolutely correct. I'd be a little nervous if we didn't have something like that.

Maybe a more manageable way to approach these tests would be to have a single, huge, annotated XML fixture that covers all of the different permutations of data and then just have one simple test scenario that just says, "when I import this huge fixture, all of the correct data should end up in the database." Managing such a fixture file might be easier than managing the test code with so many nested contexts. If the importer was less monolithic, then maybe I'd feel better about trading off some of the integration test coverage for unit test coverage.

As for refactoring the importer itself - it clearly has a lot of different responsibilities, and we're going to be piling on more. Attempting a little bit of refactoring as we go might be a good idea. That said, I don't know any particular patterns for refactoring something like this. We could move most or all of the data mapping stuff into the ActiveRecord models, I suppose  - i.e., a lot of `Publication.create_or_update_from_ai_publication` kinds of methods. Maybe extract an Activity Insight API client? Maybe add some business logic layer in between the layers that handle the mechanics of retrieving data from AI and the mechanics of persisting it to the database? The logic that will be involved in kicking off this new workflow should probably be invoked from this importer, but we might want to wrap it up in its own service object or something like that.

Those are just the best vague suggestions that I can think of - things to possibly keep in mind if you see a good opportunity to experiment with implementing any of them.

I think what you did here looks fine for getting started though.

EricDurante (approved):

PR #673, opened 2023-01-23
Adds Activity Insight Oa Workflow controller.  

### LINKED ISSUE #627: RMD admins can see a dashboard page for the new open access workflow
RMD admins will need a starting place within the RMD admin UI for any manual work that needs to be done to shepherd a publication through the new open access deposit workflow. We will probably want to create some kind of dashboard page to serve this purpose and we will likely want implement this page (along with the rest of the workflow) outside of the existing rails_admin framework/UI so that it can be completely customized and specifically tailored to this workflow.

The dashboard page doesn't need to do anything yet except enforce authentication/authorization and show some basic content.

ajkiessl:
For now it is just set up with basic authentication and admin authorization.  I had hoped I could put this behind `/admin`. However, anything behind `/admin` is caught by `RailsAdmin` where it begins looking for a model (and subsequently breaks).

Also, I added a link to this from the Rails Admin dashboard.  Controller tests for authentication/authorization as well as a test in the integration specs for the link.

closes #627

### DIFF
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -125,4 +125,10 @@ RailsAdmin.config do |config|
   end
 
   config.compact_show_view = false
+
+  # Non-rails-admin links
+  config.navigation_static_label = "Workflows"
+  config.navigation_static_links = {
+    'Activity Insight Open Access Workflow' => '/activity_insight_oa_workflow'
+  }
 end
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -127,7 +127,7 @@ RailsAdmin.config do |config|
   config.compact_show_view = false
 
   # Non-rails-admin links
-  config.navigation_static_label = "Workflows"
+  config.navigation_static_label = 'Workflows'
   config.navigation_static_links = {
     'Activity Insight Open Access Workflow' => '/activity_insight_oa_workflow'
   }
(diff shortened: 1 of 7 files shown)


config/initializers/rails_admin.rb
```diff
@@ -125,4 +125,10 @@
   end
 
   config.compact_show_view = false
+
+  # Non-rails-admin links
+  config.navigation_static_label = 'Workflows'
+  config.navigation_static_links = {
+    'Activity Insight Open Access Workflow' => '/activity_insight_oa_workflow'
```
ajkiessl, line 132: I'm not 100% on this.  It works, but it's generated by RailsAdmin so the link it creates opens the page in a new tab.  Also, you can't set the host this early in the initialization and the routes haven't been loaded, so you cannot use url helpers.

EricDurante: Yeah, it's a bummer that this isn't more configurable. I took a look, and I don't see any other reasonable way to add this to the navigation. If we wanted, we could probably add in a little javascript hack to at least make the link open in the same tab.

EricDurante (changes requested):
Can we add an integration test just to verify that this page loads when you're signed in as an admin?

ajkiessl:
> Can we add an integration test just to verify that this page loads when you're signed in as an admin?

Yup, will do.

PR #675, opened 2023-01-24
adds doi verification job and service

### LINKED ISSUE #626: Attempt to automatically verify DOI for incoming publications

Smullz622:
Adds doi verification job and service. Refactors unpaywall importer. Fixes #626


EricDurante (changes requested):
Some of the comments that I posted are just suggestions for future refactoring, but there are a couple of things that I do think we should change now - in particular, deleting any code that isn't currently used. Let me know if you have any questions. Good work!

app/jobs/doi_verification_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DoiVerificationJob
+  def perform(publication)
+    if publication.doi.present?
+      DoiVerificationService.new(publication).verify
+    else
+      response = UnpaywallClient.new.query_unpaywall(publication)
+      if publication.title&.downcase&.gsub(/[^a-z0-9]/, '') == response.title&.downcase&.gsub(/[^a-z0-9]/, '') && response.doi.present?
```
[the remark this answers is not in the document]
EricDurante: I've since noticed some of this same logic in other places. For example [here](https://github.com/psu-libraries/researcher-metadata/blob/1891128e167193b8a3b485e863f222a8d3aa7ee3/app/models/publication_merge_on_matching_policy.rb#L98) and [here](https://github.com/psu-libraries/researcher-metadata/blob/1891128e167193b8a3b485e863f222a8d3aa7ee3/app/models/publication_merge_on_matching_policy.rb#L103).

Maybe it makes sense for `Publication` and `UnpaywallResponse` to each have a method like `#matchable_title` or something of that nature..

spec/component/importers/activity_insight_importer_spec.rb
```diff
@@ -4907,9 +4914,9 @@
 
         it 'logs errors to ImporterErrorLog' do
           if import == 'Publication'
-            expect { importer.call }.to change(ImporterErrorLog, :count).by 5
+            expect { importer.call }.to change(ImporterErrorLog, :count).by 6
           else
-            expect { importer.call }.to change(ImporterErrorLog, :count).by 2
+            expect { importer.call }.to change(ImporterErrorLog, :count).by 5
```
ajkiessl, line 4919: I'm having trouble trying to understand why this went from 2 -> 5?  I can understand the one above it since we added another publication to the fixture, but I feel like this one shouldn't have changed.  I may not be considering something though.  Any ideas why this changed?

app/jobs/doi_verification_job.rb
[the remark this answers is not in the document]
Smullz622: Would it make more sense to add `#matchable_title` to `Publication` and `UnpaywallResponse` and have it just format that class's title for easy comparison or to add `#title_match?` to `Publication`, have it take in a string, handle all matching logic, and return true/false (similar to [here](https://github.com/psu-libraries/researcher-metadata/blob/1891128e167193b8a3b485e863f222a8d3aa7ee3/app/importers/unpaywall_publication_importer.rb#L137))?

spec/component/importers/activity_insight_importer_spec.rb
ajkiessl: I think maybe because an error is being raised somewhere we aren't aware of?  But the only thing that has changed is the DOIVerificationJob is being invoked.

app/jobs/doi_verification_job.rb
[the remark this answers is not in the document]
EricDurante: Good question - I'm not sure. Probably the latter if it serves all of the existing use cases.

spec/component/importers/activity_insight_importer_spec.rb
```diff
@@ -4935,19 +4942,30 @@
...
           expect(p3.activity_insight_oa_files.first.location).to eq('abc123/intellcont/file-5.pdf')
+          expect(p4.activity_insight_oa_files.first.location).to eq('abc123/intellcont/file-6.pdf')
         end
 
-        it 'does not import ActivityInsightOaFiles for imported publications without postprint/open access file locations' do
+        it 'verifies DOI for those publications' do
           importer.call
           p4 = PublicationImport.find_by(source: 'Activity Insight',
+                                         source_identifier: '171620739090').publication
+          expect(p4.doi_verified).to be true
```
ajkiessl, line 4959: I'm not too sure we should be testing how the DOIVerificationJob works in this test since it is an async job.  Perhaps we should test that the job is queued or that `DoiVerificationJob.new.perform(pub_record)` was invoked instead.

ajkiessl: I think it may be prone to breaking if we rely on the outcome of an async job.

spec/component/importers/activity_insight_importer_spec.rb
EricDurante: Good catch, @ajkiessl! I didn't notice this. Is there a corresponding decrease in the number of records being successfully imported somewhere?

spec/component/importers/activity_insight_importer_spec.rb
EricDurante: Yeah, we might end up just removing the job queuing from the importer though, right?

spec/component/importers/activity_insight_importer_spec.rb
ajkiessl: I popped a byebug in the error rescue for the publication part of the import and the error is a `<JSON::ParserError: unexpected token at ''>`.  It's being risen from `DoiVerificationJob.new.perform(pub_record)`.  Not sure if that's from something inside the job, or if delayed job is failing for some other reason.  I thought these jobs would be queued and ran in a separate process? But maybe delayed job works differently in the test environment and runs the process synchronously.

Smullz622: It seems like it's a problem with the spec and how the HttpService is mocked. Eric & I met, and we're going to just mock `DoiVerificationJob` instead of trying to return an appropriate response for each http request. Then we can also use that to check that the job was called (for now until we decide when/if we want to remove it from the importer)

ajkiessl (approved):
This looks good to me.  Thanks and nice work!

EricDurante (approved):

PR #682, opened 2023-02-01
DOI Verification Dashboard

ajkiessl:
Screen Shot 2023-02-01 at 1 12 20 PM

☝️ The "bucket" that links to the list of publications that need their DOIs checked.

Screen Shot 2023-02-01 at 1 12 32 PM

☝️ The list of publications that need their DOIs checked.

This list will only contain publications with `doi_verified` fields that are set to `false`.  Any that are `nil` will not show here.  Those will be left to be picked up by the cronjob.  Note, this doesn't contain links to edit the publications yet.  That will be addressed in a separate issue.

I should also add that this list is not paginated.  So if there are 1000 records, then 1000 records display.  This may be something we want to consider in the future, but I don't think we'll have enough records here that it _needs_ pagination immediately.


ajkiessl:
https://github.com/psu-libraries/researcher-metadata/pull/683

PR #683, opened 2023-02-01
DOI Verification Dashboard

ajkiessl:
Screen Shot 2023-02-01 at 1 12 20 PM

☝️ The "bucket" that links to the list of publications that need their DOIs checked.

Screen Shot 2023-02-01 at 4 16 12 PM

☝️ The list of publications that need their DOIs checked.

This list will only contain publications with `doi_verified` fields that are set to `false`.  Any that are `nil` will not show here.  Those will be left to be picked up by the cronjob. 

I should also add that this list is not paginated.  So if there are 1000 records, then 1000 records display.  This may be something we want to consider in the future, but I don't think we'll have enough records here that it _needs_ pagination immediately.


app/models/publication.rb
```diff
@@ -132,6 +132,10 @@ def self.open_access_statuses
   scope :oa_publication, -> { where(publication_type: oa_publication_types) }
   scope :non_oa_publication, -> { where.not(publication_type: oa_publication_types) }
 
+  scope :not_open_access, -> { distinct(:id).left_outer_joins(:open_access_locations).where(open_access_locations: { publication_id: nil }) }
```
EricDurante, line 135: Given the larger context, this looks more like "unknown_access" than "not_open_access" to me, although that's not quite right either because there isn't enough info here to say for sure that the access is unknown. I think all that we can really determine from this query is that these are publications that don't yet have any open access locations that we know about. I'd read "not_open_access" and immediately think `open_access_status = 'closed'`.

Maybe just "no_oa_locations" is a descriptive enough name for this scope without hinting at any extra meaning that isn't really there.

Naming aside, would we want to include any publications in this workflow based on the value of the `open_access_status` field? (This isn't a rhetorical question - I'm really asking.) For example, if there's a publication that has a known open access location, but we know that the open access location might only be temporary because we also know that the open_access_status is 'bronze', then would we still want to try deposit the file that we've been given?

app/models/publication.rb
```diff
@@ -488,6 +494,12 @@ def doi_url_path
       field(:edition)
       field(:page_range)
       field(:doi) { label 'DOI' }
+      field(:doi_verified, :enum) do
+        label 'DOI verified?'
+        enum do
+          [['True', true], ['False', false]]
+        end
+      end
```
EricDurante, line 502: I'm assuming that leaving the field blank (null) here is an option too and that's why you used an enum instead of a checkbox, correct?

spec/component/models/publication_spec.rb
```diff
@@ -460,6 +460,51 @@
...
+    let!(:pub3) { create(:publication,
+                         title: 'pub3',
+                         doi_verified: true)
+    }
+    let!(:pub4) { create(:publication,
+                         title: 'pub4',
+                         doi_verified: nil)
+    }
+    let!(:pub5) { create(:publication,
+                         title: 'pub4',
```
EricDurante, line 480: Is this title a typo?

spec/component/models/publication_spec.rb
```diff
@@ -460,6 +460,51 @@
...
+    let!(:activity_insight_oa_file1) { create(:activity_insight_oa_file, publication: pub2) }
+    let!(:activity_insight_oa_file2) { create(:activity_insight_oa_file, publication: pub3) }
+    let!(:activity_insight_oa_file3) { create(:activity_insight_oa_file, publication: pub4) }
+
+    let!(:open_access_location) { create(:open_access_location, publication: pub5) }
+
+    describe '.not_open_access' do
+      it 'returns publications that do not have open access information' do
+        expect(described_class.not_open_access.map(&:title)).to match_array [pub1, pub2, pub3, pub4].map(&:title)
+      end
```
EricDurante, line 492: It's a little hard for me to tell what's going on here given that there's a publication with a duplicate title and we're comparing result sets by title. Curious what your reasoning is for doing it that way..

EricDurante (commented):
This looks great. I'm not sure that anything needs to be changed - just have some questions.

app/models/publication.rb
ajkiessl: Correct.  Also, if an admin were to make any edits to a publication with a doi_verified as `nil`, the boolean would change from `nil` to `false` on update, which could throw things off.

spec/component/models/publication_spec.rb
ajkiessl: Yeah, that's a typo.

spec/component/models/publication_spec.rb
ajkiessl: The intention was to exclude pub5 since it has an open access location, but the typo is making this not quite make sense.  I don't really have a good reason for only matching the titles other than that's how the other scopes are tested above.

EricDurante: Ok, cool - if it'll work fine, then I'd match on the objects/records themselves instead of their titles. I'd definitely find that to be less confusing since that's really what we want to assert here.

app/models/publication.rb
ajkiessl: I'll change the name of that scope to be more specific.

For the second part, maybe.  I'm not too sure.  It may be something we want to confirm with Ana later, and consider adding in the future.

EricDurante (approved):
Excellent!

PR #685, opened 2023-02-02
Adds oa workflow state to publication

### LINKED ISSUE #680: Add OA Workflow State field to Publication
We'll want to store state information about where in the OA workflow a publication is.  For now, we want to know when a publication has stopped for manual curation, or when the regular cron job to trigger next steps has queued a publication.  

We'll want to store a constant in the publication model that will hold the allowable states in an array.  We should then use this to validate that a state being set is valid.

Smullz622:
Fixes #680


EricDurante (approved):

ajkiessl (dismissed):
LGTM

ajkiessl (approved):

EricDurante (approved):

PR #687, opened 2023-02-03
adds oa workflow rake task and service

### LINKED ISSUE #679: OA Workflow CronJob
We'll need a cronjob that runs fairly regularly (as often as possible without overlapping), that will trigger next steps in the workflow.  

We are no longer going to try to build this workflow linearly, with each step triggering the next after completion.  Since there are too many cases where data could be changed elsewhere in RMD (knocking things out of place in the workflow), it makes more sense to have a regular process that looks for criteria (or scopes) to trigger next steps (jobs) for publications in this workflow.

Initially, we'll set this up to handle the DOI check, since it is the first one we have completed.  We'll need separate tickets to implement to other steps of the workflow.

Smullz622:
fixes #679

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -140,8 +140,7 @@ class Publication < ApplicationRecord
   scope :with_no_oa_locations, -> { distinct(:id).left_outer_joins(:open_access_locations).where(open_access_locations: { publication_id: nil }) }
   scope :activity_insight_oa_publication, -> { with_no_oa_locations.joins(:activity_insight_oa_files).where.not(activity_insight_oa_files: { location: nil }) }
   scope :doi_failed_verification, -> { activity_insight_oa_publication.where('doi_verified = false') }
-  scope :doi_nil_verification, -> { activity_insight_oa_publication.where(doi_verified: nil) }
-  scope :needs_doi_verification, -> { doi_nil_verification.where.not(oa_workflow_state: 'automatic DOI verification pending').or(doi_nil_verification.where(oa_workflow_state: nil)) }
+  scope :needs_doi_verification, -> { activity_insight_oa_publication.where(doi_verified: nil).where(%{oa_workflow_state IS DISTINCT FROM 'automatic DOI verification pending'}) }
 
   scope :published, -> { where(publications: { status: PUBLISHED_STATUS }) }
 
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -140,6 +140,7 @@ class Publication < ApplicationRecord
   scope :with_no_oa_locations, -> { distinct(:id).left_outer_joins(:open_access_locations).where(open_access_locations: { publication_id: nil }) }
   scope :activity_insight_oa_publication, -> { with_no_oa_locations.joins(:activity_insight_oa_files).where.not(activity_insight_oa_files: { location: nil }) }
   scope :doi_failed_verification, -> { activity_insight_oa_publication.where('doi_verified = false') }
+  scope :needs_doi_verification, -> { activity_insight_oa_publication.where(doi_verified: nil) }
 
   scope :published, -> { where(publications: { status: PUBLISHED_STATUS }) }
 
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -65,7 +65,7 @@ class Publication < ApplicationRecord
   end
 
   def self.oa_workflow_states
-    ['queued for doi']
+    ['automatic DOI verification pending']
   end
 
   has_many :authorships, inverse_of: :publication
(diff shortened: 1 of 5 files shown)


app/services/oa_workflow_service.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class OaWorkflowService
+  def workflow
+    Publication.needs_doi_verification.each do |pub|
```
EricDurante, line 5: Here's where I think we're going to need to set the publication's `oa_workflow_status` to be something like "automatic DOI verification pending" and then exclude publications with that status from the `needs_doi_verification` scope on `Publication`. Otherwise, depending on how often the cron job runs and how long it takes for `DoiVerificationJob` to attempt to verify a DOI, we could end up queuing up (and running) more instances of the same job for the same publication before the first one has a chance to finish.

I think we may as well go ahead and do that now.

EricDurante (changes requested):
Everything looks good! I just think this is a good point to go ahead and add the constraint to prevent duplicate DOI verification jobs from potentially piling up.

app/services/oa_workflow_service.rb
ajkiessl: I think it may be possible to lose a publication if an error occurs after the `oa_workflow_status` is set and the doi_check ends up staying nil.  This may be something that almost never happens, since we already handle error in the client class, but I think it could still happen?  Or perhaps we want the error to stop the publication from proceeding/checking again, so we can address the error.  I'm actually not too sure about this one.

Smullz622: Current plan is to still set the `oa_workflow_status` at the beginning but add a rescue block that will set `doi_verified` to false. This will kick it over for manual verification and keep it from getting queued for the job again.

Smullz622:
In publication, the `needs_doi_verification` scope wasn't handling the nil well for `oa_workflow_state`. It was fixed by adding an `.or` to check if `oa_workflow_state` is nil. I added another scope for `doi_nil_verification` to clean up the code.

app/models/publication.rb
```diff
@@ -140,6 +140,8 @@ def self.oa_workflow_states
   scope :with_no_oa_locations, -> { distinct(:id).left_outer_joins(:open_access_locations).where(open_access_locations: { publication_id: nil }) }
   scope :activity_insight_oa_publication, -> { with_no_oa_locations.joins(:activity_insight_oa_files).where.not(activity_insight_oa_files: { location: nil }) }
   scope :doi_failed_verification, -> { activity_insight_oa_publication.where('doi_verified = false') }
+  scope :doi_nil_verification, -> { activity_insight_oa_publication.where(doi_verified: nil) }
+  scope :needs_doi_verification, -> { doi_nil_verification.where.not(oa_workflow_state: 'automatic DOI verification pending').or(doi_nil_verification.where(oa_workflow_state: nil)) }
```
EricDurante, line 144: @ajkiessl @Smullz622 
Do we actually need the `oa_workflow_state: nil` condition here?

A `nil` value for `doi_verified` is what tells us that the job needs to be run, and the only exception to that rule that I can think of is if `oa_workflow_state` is `"automatic DOI verification pending"`. Isn't it ok if `oa_workflow_state` is any value other than `"automatic DOI verification pending"`?

In other words, is there a problem with
```ruby
scope :needs_doi_verification, -> { activity_insight_oa_publication.where(doi_verified: nil).where.not(oa_workflow_state: 'automatic DOI verification pending') }
```
that I'm not considering?

Smullz622: That's what I had at first, but it wouldn't pick up publications where `oa_workflow_state` was nil. It only pulled ones that had a string & that string was not `"automatic DOI verification pending"`. Once I added the `oa_workflow_state: nil` condition, it pulled the appropriate publications.

app/services/oa_workflow_service.rb
```diff
@@ -0,0 +1,14 @@
...
+      pub.oa_workflow_state = 'automatic DOI verification pending'
+      pub.save!
+      DoiVerificationJob.new.perform(pub)
+    rescue StandardError
+      pub.doi_verified = false
+      pub.save!
```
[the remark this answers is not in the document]
ajkiessl: I agree, I think we could just raise the error again at the end of the rescue block.  Then, we'll be able to see it in Datadog and potentially fix it.

app/models/publication.rb
ajkiessl: Iirc this is what we were confused with yesterday.  The scope you have above wouldn't include publications with `oa_workflow_state: nil`.  Not sure why, but it wasn't working in the tests.

EricDurante: ohhh right. Generating the SQL for the scope that I suggested yields
```sql
SELECT DISTINCT "publications".* FROM "publications" INNER JOIN "activity_insight_oa_files" ON "activity_insight_oa_files"."publication_id" = "publications"."id" LEFT OUTER JOIN "open_access_locations" ON "open_access_locations"."publication_id" = "publications"."id" WHERE "open_access_locations"."publication_id" IS NULL AND "activity_insight_oa_files"."location" IS NOT NULL AND "publications"."doi_verified" IS NULL AND "publications"."oa_workflow_state" != 'automatic DOI verification pending'
```
with ActiveRecord's `.not` translating to the SQL comparison operator `!=` for `oa_workflow_state` which returns `NULL` instead of `true` when a record's `oa_workflow_state` is `NULL`.

So using the ActiveRecord `.or` works, but (and this is just my opinion) I would find it easier to read and understand if it was all in one scope (since I don't think that we need the intermediate `doi_nil_verification` scope for anything else) and written in raw SQL. So then you could do the SQL `OR` in combination with the `AND`, or maybe even better - you could use a different method of comparison:
```ruby
scope :needs_doi_verification, -> { activity_insight_oa_publication.where(doi_verified: nil).where(%{oa_workflow_state IS DISTINCT FROM 'automatic DOI verification pending'}) }
```
I think that does the trick.

EricDurante (approved):
Looks good!

ajkiessl (approved):

PR #694, opened 2023-02-09
Sort duplicate publication groups from lowest to highest id

ajkiessl:
Don't reverse sort duplicate publication groups so that the oldest are first in the list.  Also, running db migrations removed a space in the schema so there's that too.


Smullz622 (approved):
lgtm

banukutlu (approved):
👍🏻

PR #695, opened 2023-02-09
adds fetch oa status job

### LINKED ISSUE #686: Automatically fetch open access status/info for incoming publications with confirmed DOI
We already have cron jobs that import open access metadata from Open Access Button and Unpaywall for all of the publications in RMD on a regular basis. However, these jobs take a long time to run, and they run relatively infrequently. For the small subset of publications that will be entering this workflow, we want to fetch this information much more quickly at the outset so that the publications can proceed through the workflow as rapidly as possible.

So, when the main cron job that orchestrates this workflow finds a publication with a file upload from Activity Insight and a confirmed DOI but no open access metadata, it should enqueue a background job to fetch the open access metadata from Open Access Button and/or Unpaywall based on the DOI. We should do any refactoring that is necessary to share as much code as possible between this background job and the existing OAB and Unpaywall importers.

Just before enqueueing the job, we'll need to change the open access workflow status of the publication to indicate that this automated step is pending. This will require us to add a new value to the list of possible values for a publication's open access workflow status and then to filter out publications with that status when enqueuing jobs for this step. If the job completes successfully and records open access metadata, then we can clear this status as the last step.

If the automated search for open access metadata fails (i.e. because OAB and Unpaywall respond with a 404 given a publication's DOI), then we'll need to be able to set another new open access workflow status on the publication to indicate that an RMD admin needs to do a manual check for open access information for the publication.

Smullz622:
Fixes #686 

Adds FetchOAStatusJob to OaWorkflowService
Refactors OAButton and Unpaywall importers


app/models/publication.rb
```diff
@@ -141,6 +147,7 @@ def self.oa_workflow_states
   scope :activity_insight_oa_publication, -> { with_no_oa_locations.joins(:activity_insight_oa_files).where.not(activity_insight_oa_files: { location: nil }) }
   scope :doi_failed_verification, -> { activity_insight_oa_publication.where('doi_verified = false') }
   scope :needs_doi_verification, -> { activity_insight_oa_publication.where(doi_verified: nil).where(%{oa_workflow_state IS DISTINCT FROM 'automatic DOI verification pending'}) }
+  scope :needs_oa_metadata_search, -> { activity_insight_oa_publication.where(doi_verified: true).where(%{oa_workflow_state IS DISTINCT FROM 'no open access data found'}) }
```
EricDurante, line 150: Depending on what we decide regarding my earlier comment, we might be doing away with `'no open access data found'` and either replacing it with something else, or not replacing it at all.

However, in any case, I think we do need to exclude publications with an `oa_workflow_state` of `'oa metadata search pending'`.

app/jobs/fetch_oa_metadata_job.rb
```diff
@@ -0,0 +1,18 @@
...
+    oab_response = OABClient.query_open_access_button(publication)
+
+    if unpaywall_response.oa_locations.present?
+      publication.update_from_unpaywall(unpaywall_response)
+    elsif oab_response.url
+      publication.open_access_locations.create!(source: Source::OPEN_ACCESS_BUTTON, url: oab_response.url)
+      publication.open_access_button_last_checked_at = Time.zone.now
+    end
+
+    publication.oa_workflow_state = (publication.open_access_locations.present? ? 'open access publication' : 'no open access data found')
```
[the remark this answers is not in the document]
EricDurante: Depending on how we decide to make this work, we might need to clear the `'oa metadata search pending'` value in `oa_workflow_status` when the job completes.

EricDurante (commented):
@Smullz622 Again, all of the refactoring that you did here is great. I think it's really a big improvement. I'm sorry for the churn on how the feature should actually work. I hadn't quite thought this issue through well enough. I think that I know what we should do now, but I'd like to have @ajkiessl 's input. Fortunately, whatever we decide, I don't think that  it will result in a large amount of change.

app/jobs/fetch_oa_metadata_job.rb
```diff
@@ -0,0 +1,18 @@
...
+    oab_response = OABClient.query_open_access_button(publication)
+
+    if unpaywall_response.oa_locations.present?
+      publication.update_from_unpaywall(unpaywall_response)
+    elsif oab_response.url
+      publication.open_access_locations.create!(source: Source::OPEN_ACCESS_BUTTON, url: oab_response.url)
+      publication.open_access_button_last_checked_at = Time.zone.now
+    end
+
+    publication.oa_status_last_checked_at = Time.zone.now
```
EricDurante, line 15: Here I think we'll also need to clear the publication's `oa_workflow_state`. It will have been set to `"oa metadata search pending"` prior to enqueuing the job, which will exclude it from any subsequent checks unless we set `oa_workflow_state` back to `nil` upon completion of the job. Does that sound right?

app/models/publication.rb
```diff
@@ -65,7 +65,13 @@ def self.open_access_statuses
   end
 
   def self.oa_workflow_states
-    ['automatic DOI verification pending']
+    [
+      'automatic DOI verification pending',
+      'oa metadata search pending',
+      'error during oa metadata search',
+      'open access publication',
+      'no open access data found'
```
EricDurante, line 73: I think we can get rid of `'open access publication'` and `'no open access data found'` now, correct? `'open access publication'` will instead be indicated by the presence of `OpenAccessLocation`s, and `'no open access data found'` isn't a state that we care about any more since we'll be running the check continuously regardless.

I sort of don't think that we need to record `'error during oa metadata search'` either since we're currently not doing anything with that information. If an error does occur and the cause is transient, then active_job should retry the job until it succeeds. And even if none of those retries succeed, it'll be queued up again an hour later. So we'd only need to know about really persistent errors, and I think we can find out about those just by allowing the job to blow up without rescuing anything at all.

EricDurante (changes requested):

ajkiessl (approved):
This all looks good to me.  Thanks for the refactoring.  The open access location importers look much better with these changes.

EricDurante (approved):
Looks good - thanks!

PR #699, opened 2023-02-15
Record failure to find missing DOI

EricDurante:
I went back and forth on how to accomplish the behavior that I think that want here. Ultimately I think that the list should continue to behave the way that it currently does and only show publications with `doi_verified: false`, and the DOI verification job should set the flag to false if the publication doesn't have a DOI and the job couldn't find one - that way, every publication that comes out of the automated DOI check should have the verified flag set to either true or false.


ajkiessl:
Yup, I agree.  This looks good, I just pushed up some rubocop fixes.  I'll merge when CI passes, thanks!

ajkiessl (approved):

PR #711, opened 2023-02-22
adds specs for doi verified hot fix

Smullz622:


EricDurante (dismissed):
:shipit:

ajkiessl:
Something strange has been going on with the `spec/fixtures/vcr_cassettes/PSUIdentityUserService` directory since I changed it from `PsuIdentityUserService` to `PSUIdentityUserService`.  They get deleted for me as well sometimes.  Idk if they're not being tracked properly by git?

Smullz622:
I think one of the contexts was changed slightly & that threw them all off since they are custom. I just redid them & think it should be good now.

EricDurante (approved):

PR #712, opened 2023-02-22
limits OA workflow to open access publication types

### LINKED ISSUE #709: Limit OA Workflow to publication types that are subject to OA policies
Our `activity_insight_oa_publication` base scope in `Publication` should be limited to the `oa_publication` scope (publication types that are subject to OA policies).

We also don't want to import files for publications that aren't subject to OA policies, so we'll want to apply some logic to the Activity Insight importer that only imports files for OA publications.

Smullz622:
Fixes #709


EricDurante (approved):
Looks good to me!

PR #713, opened 2023-02-22
File version checking delayed job

### LINKED ISSUE #575: File Version pattern checking
We want to implement version checking in a similar way OA Works does.

OA Works uses this spreadsheet of regex patterns to help them make a determination if a file is an accepted version (post-print) or published version.  They apply these patterns to check the first 500 words of a document and the filename.  The spreadsheet indicates what pattern to look for and what version it indicates.  The spreadsheet can be found here: https://docs.google.com/spreadsheets/d/10DNDmOG19shNnuw6cwtCpK-sBnexRCCtD4WnxJx_DPQ/edit#gid=0 .

Their code basically looks for the patterns and adds a point if a published version pattern is found and removes a point if an accepted manuscript pattern is found.  If the end score is greater than 0, it is a published version, if it is less than 0 it is an accepted manuscript.  When talking with someone from OA Works it sounded like they weighed published version patterns a little more heavily.  We'll have to take a look at some of these patterns and see if that's what we want to do.  OA Work's code can be found here: https://github.com/oaworks/api/blob/master/worker/src/deposit.coffee#L260-L297 . 

OA Works indicated that this processing takes 15-30 seconds, so we may want to add a loading animation to the submit files button so users know it is processing.

banukutlu:
Closes #575 

TODO:

- [x] Refactor File version check as a background job possibly in another PR.
- [x] If max polling reached, display the partial with we could not find a file version text and radio-buttons
- [x] Refactor ScholarspherePdfFileVersion and extract a couple of other classes from the private behavior which will also make unit testing easier.


app/controllers/open_access_publications_controller.rb
```diff
@@ -57,6 +69,48 @@ def scholarsphere_file_version
...
+          false
+        else
+          job.destroy
+          Rails.logger.error "Job with ID #{job_id} failed and has been removed"
+          true
+        end
+      else
+        false
+      end
+    end
```
EricDurante, line 93: I don't know how important it is that this bit happens, but if I'm understanding everything correctly, then it seems like this is dependent on the client polling for results of the file version check. So in other words, if the client doesn't poll (i.e. the user closes their browser or navigates away from the page), then this failed job cleanup and error logging doesn't occur. Is that correct? If so, would that be a problem?

app/views/open_access_publications/scholarsphere_file_version.html.erb
EricDurante, line 1: I'm curious what happens when the user reloads/refreshes this page both before and after the javascript polling has returned a result. Does it show the correct state in both cases? Does reloading the page in either case have any other consequences?

I'd guess that this is something that users are likely to do, so it would be good to make sure that we've tried it and verified that it's well-behaved.

app/jobs/scholarsphere_pdf_file_version_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ScholarspherePdfFileVersionJob < ApplicationJob
+  queue_as 'scholarsphere-pdf-file-version'
+
+  def perform(file_meta:, publication_meta:, exif_file_version:)
+    pdf_file_version = ScholarspherePdfFileVersion.new(file_meta: file_meta, publication_meta: publication_meta).version
+
+    Rails.cache.write("file_version_job_#{job_id}", { pdf_file_version: pdf_file_version, file_meta: file_meta, exif_file_version: exif_file_version })
```
EricDurante, line 9: Admittedly, I don't have my head completely wrapped around everything that's going on in this PR, so maybe the answer is actually obvious, but I'm curious about what led to the decision to use Rails caching to store this information vs. using the database.

Also, what cache store are we planning to use in production?

app/models/scholarsphere_pdf_file_version.rb
EricDurante, line 1: No need to take action now, but I'd keep this class in mind as a possible target for future refactoring. There's a whole lot of private logic behind a single public method, which partly hides the fact that this class has many disparate responsibilities. At some point, I think that it might be worthwhile to extract a couple of other classes from the private behavior here - even if it's for no other reason than to make thorough unit testing easier.

app/controllers/open_access_publications_controller.rb
```diff
@@ -36,16 +36,28 @@ def update
...
+      @cache_files = file_version_uploads.cache_files
+      exif_versions = file_version_uploads.exif_file_versions.compact
+      file_version = ScholarsphereFileVersionUploads.version(exif_versions) if exif_versions.present?
+
+      if file_version == I18n.t('file_versions.published_version')
+        render partial: 'open_access_publications/file_version_result', locals: { file_version: file_version, cache_files: @cache_files }
+      else
+        @cache_files.each do |cache_file|
+          file_version_job = ScholarspherePdfFileVersionJob.perform_later(file_meta: cache_file.to_json, publication_meta: version_check_pub_meta, exif_file_version: file_version)
+          @job_ids.push(file_version_job.job_id)
```
EricDurante, line 54: I'm also curious about the need to pass around an array of job ID's between HTTP requests in order to check on the status of the jobs as opposed to polling to check on the state resulting from the jobs having been run. In other words, how come we're watching background jobs rather than, for example, checking the state of file upload records in the database?

I'm not saying that this wasn't necessary - I'm really just asking because I don't have quite enough context to understand the reasoning for doing it this way.

app/models/scholarsphere_pdf_file_version.rb
banukutlu: Thanks @EricDurante for your feedback. Agreed there is so much private logic going on here. I will add it as todo.

app/jobs/scholarsphere_pdf_file_version_job.rb
banukutlu: Currently, we are planning to use file caching in production for this work. However, since this is still a work in progress and is being tested on a preview branch, it will not be deployed just yet. We plan to revisit this in the next cycle.

config/file_version_checking_rules.csv
EricDurante, line 1: Yeah, even though it feels a little awkward, I think that processing this CSV file is definitely the right way to go. For better or worse, this file essentially _is_ the API for OA.Works' version checking. They can modify/extend/improve their process just by changing the rules in this file, and as long as they don't change the logic for how the file is processed, then we can benefit from their continued work. If they published this file on the web somewhere, we could even fetch it dynamically (and probably cache it for some period of time) to get updates automatically. We'd only want to do that if they could guarantee that there wouldn't publish any breaking changes though.

app/controllers/open_access_publications_controller.rb
ajkiessl: @banukutlu correct me if I'm wrong, but this job doesn't result in anything changing in the database.  It's just used to prefill the form on the next page.  

The Scholarsphere upload process is now a three-step workflow.  The final step creates the upload record.  The version checking is between the first and second step.

banukutlu: > @banukutlu correct me if I'm wrong, but this job doesn't result in anything changing in the database. It's just used to prefill the form on the next page.
> 
> The Scholarsphere upload process is now a three-step workflow. The final step creates the upload record. The version checking is between the first and second step.

Yes, that's correct. It doesn't modify the database or store any results in a field. Instead, it simply populates a form with the file version information, but only if a result is found through the version check. Ultimately, it's up to the user to decide whether to accept the prefilled value or make changes.

banukutlu: > I'm also curious about the need to pass around an array of job ID's between HTTP requests in order to check on the status of the jobs as opposed to polling to check on the state resulting from the jobs having been run. In other words, how come we're watching background jobs rather than, for example, checking the state of file upload records in the database (or where ever the jobs' results are being stored)?
> 
Eric, as Alex mentioned, we don't store the status of file version checks in the database. We pass an array of job IDs to keep track of the jobs we need to monitor. We're open to any ideas you may have on this.

app/controllers/open_access_publications_controller.rb
banukutlu: I added this part while writing the specs, but you are right, it seems that I made a wrong assumption. I will be revising this section. Thank you for bringing it to my attention.

app/jobs/scholarsphere_pdf_file_version_job.rb
whereismyjetpack: The more I think about this, the more I think my ignorance of rmd's data model may be getting in the way 

Perhaps we could add a `version` column to the work deposit, and use that as an indicator? 

I think that the reason we don't do this is because a work can have many files, and we need to make sure we've processed all the files before determining the final version. 

I'd be open to discuss this sometime soon, because I do think there's room for improvement perhaps.

app/controllers/open_access_publications_controller.rb
banukutlu: To answer your question about the cases if user closes browser and logging does not happen, that would not be a problem esp since it is not saving anything to database.

app/jobs/scholarsphere_pdf_file_version_job.rb
ajkiessl: There's a couple reasons we don't use a version column on the work deposit:

- The version is ultimately determined by the user.  The automated version checking just prefills the form on the next page, so the user can confirm.
- The work deposit isn't actually created until the end of the workflow.  This could be changed, but then we'd have to come up with a mechanism for removing or restarting the created work deposit if a user leaves the workflow without finishing.

ajkiessl: Not saying we couldn't do it this way, but it might take a little (or a lot) of extra work.

app/controllers/open_access_publications_controller.rb
EricDurante: That makes sense. Thanks for the explanation!

config/file_version_checking_rules.csv
banukutlu: yes, I agree!

ajkiessl:
I didn't quite follow the above TODOS exactly.  Also, I changed a decent amount of other things from the original PR that Banu put together.  

I changed how the controllers were handling the exif version checking results vs the new version checking results.  I think we were overthinking things and over-complicating by trying to pass the exif result through the process, then try to figure out which is better at the end (exif or non-exif).  The way it works now, if the exif checker finds something, it goes with that.  Otherwise, it does the new version checker.

I also changed how we were giving preference when multiple files were uploaded.  I added `score` to the version checker interface so we can compare absolute scores between files and use the highest value.

I made sure the FileVersionChecker can easily be used over in with the OA Workflow.  We should be able to pass the file_path and publication to that class from the OA Workflow context.  We'll want to create a separate job for those, though, so they get sent to a different queue.

I added the 60 second timeout.

Rather than refactor FileVersionChecker to make unit testing easier, I made some integration tests that cover the extent of the class's logic.

I added a bunch of error handling.

EricDurante (approved):
I'll try to look this over more closely if I get a chance, but I don't want it to be held up waiting for my approval. I'm pretty far out of my element on this one, so I'm not sure whether or not I'm going to have anything else of value to add unless I spend a _long_ time really trying to understand everything that's going on here.

PR #718, opened 2023-02-27
File version review list

### LINKED ISSUE #688: "Bucket" and list for publications that need manual file version checking
We'll need another card on the main OA Workfow dashboard that describes the manual file version checking step and has the number of publications in the bucket in the top right corner.  This card should also be a link to a list of publications that need manual file version checking.  

We don't currently have a field to store file version on the `ActivityInsightOaFile` model, so that will need to be added.  This should be restricted to published and accepted versions kind of like how we do it in `OabPermissionsService`.  We'll need a scope to select the publications that will be in this bucket.  We'll also need another status added to the `oa_workflow_statuses` that this scope will use to select publications for this bucket. 

The list should have the title of the publication the file belongs to, and the file location.

ajkiessl:
This list contains any publications that have a preferred version and all of its Activity Insight OA file versions are either 'unknown' or their version does not match up with the preferred version.  The list looks like this:

Screen Shot 2023-02-27 at 3 58 43 PM

And the card in the main dashboard 👇 

Screen Shot 2023-02-27 at 4 00 52 PM

This is just the list.  The editing portion will be handled in another issue.

closes #688


ajkiessl:
I need to modify the scope for this list.  It still includes publications whose preferred_version and all of it's files' versions are determined, but the file versions are all wrong.  The plan is to put these in a separate list.

ajkiessl:
> I need to modify the scope for this list. It still includes publications whose preferred_version and all of its files' versions are determined, but the file versions are all wrong. The plan is to put these in a separate list.

Done

ajkiessl:
I'm noticing that we are requiring the file version to be either 'acceptedVersion' or 'publishedVersion' to fit the scope for the permissions check:
https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/activity_insight_oa_file.rb#L8
This will conflict with the scope for the list in this PR since the list requires a preferred_version.  There will be instances where all the files are 'unknown' and we will not have a preferred_version.  I could remove the requirement in the 'needs_permissions_check` that the version is either 'acceptedVersion' or 'publishedVersion', or I could remove the requirement for a preferred_version in this list's scope.  My thought was the preferred_version could help admins determine the version for the unknowns in this list, so I'm thinking of doing the former solution.  @EricDurante @Smullz622 Does that sound ok?  Or am I missing something?

Smullz622:
@ajkiessl My inclination would be to remove it from the `needs_permissions_check` scope so we can retrieve preferred version regardless, but keep the conditional logic around the other permissions. If preferred is publishedVersion but we have an acceptedVersion and we save permissions data, we'll get the wrong licence, etc.

Smullz622:
Unrelated, but just thought of it- we'll probably also want to update the `version_checked` column in file to be more specific (maybe `version_permissions_checked`) so we don't get it confused once we add the actual version checking job. My bad on that one.

ajkiessl:
> If preferred is publishedVersion but we have an acceptedVersion and we save permissions data, we'll get the wrong licence, etc. If an admin manually reviews and decides acceptedVersion is fine to deposit, will they know to delete the incorrect data?

Yeah I can see this being an issue in the cases where it may be fine for the author to deposit either type of version.  I guess in this case an admin could manually change the `preferred_version`.  Maybe the best solution right now would be to move that `version_checked` update into the `PermissionsCheckJob` and only update when the permissions data is set (when we have a file version that matches the preferred version).  If an admin does ever change the `preferred_version` for any reason, though, they should still be aware that the permissions data may need to be updated. 

> Unrelated, but just thought of it- we'll probably also want to update the version_checked column in file to be more specific (maybe version_permissions_checked) so we don't get it confused once we add the actual version checking job. My bad on that one.

I agree we should probably change this.

ajkiessl:
Or perhaps we could just expand the scope here:
`.where(%{version = ? OR version = ?}, I18n.t('file_versions.accepted_version'), I18n.t('file_versions.published_version'))`

To include files whose version is not null.  So we would still get a `preferred_version` even when the file version is 'unknown'.  We wouldn't get permissions data though, even if the 'unknown' was changed to be the `preferred_version`.

Smullz622:
Another thought, would it make any sense to use `OABPreferredPermissionsService` outside of the current workflow step/scope to retrieve `preferred_version` when file version is unknown? Then `version_checked` (or whatever we rename it) wouldn't interfere and it could still be rerun later to get permissions data

ajkiessl:
It might be difficult to do that.  We could trigger the service in the setter for `ActivityInsightOAFile`'s `version` attribute.  That could get problematic though, since it would trigger any time that attribute is updated.  I'm not sure there's any other place in the workflow you could hook in the service.

EricDurante:
Something about this interplay between the file version checking and the permissions checking feels like it's more complicated than it needs to be, and I might be partly to blame for that over-complication.

I had originally thought that we would want to first determine the version of a file and then try to find out the permissions for that file/version. There were two assumptions baked into that. The first assumption was that we'd be querying for permissions both by DOI and version, and then getting back the permissions just for that version. The other assumption was that we'd get permissions for whatever file versions that we had downloaded and then store that as metadata for each of the files. So eventually we'd end up with a bunch of files where we'd know the version of each one and we'd know the permissions for each one based on the version. Then to determine whether or not we had a file that could be deposited, we could query for a file that had permissions information indicating that it can be deposited into an institutional repository. Given these assumptions, it made sense to do permissions checking per file and only after we had done the version checking for that file.

That's not how it's shaking out though because my assumptions were wrong. We don't query for permission metadata by DOI and version - we query only by DOI, and we get back all of the permissions metadata for every version of the publication. We're also not storing permissions metadata per file even though the individual permissions do pertain to a file with a version rather than to the publication as a whole. That part actually makes perfect sense though, I think. We don't need to store the permissions metadata for _every_ file/version that we have - we just need to store the information telling us which version of the publication we can deposit and what the conditions of deposit (set statement, embargo date, etc.) are for that version. It makes total sense to store that metadata with the publication record, and I do think that we probably only need to try that automated check/import once (at least for now).

So now that I understand the lay of the land better, I think that permissions checking should actually happen per _publication_ rather than per file, and it doesn't need to be dependent on any file version checking happening first. It only needs the DOI to be verified first. The file version checking has no prerequisites at all apart from downloading the file first.

Once the permissions check has been done, the publication record can tell us what version can be deposited, and once the version checking is done, we can query for a file with that version. Once that query gives us a result, the publication is ready for the last step/list for metadata review and deposit. So that's the scope for that last list - publications that have a file with a  (non-null) version that matches the (non-null) version stored in the publication. Our list for manual file version checking can exclude any publications (or files that belong to publications) that satisfy that scope and appear in the final list, and it can include any remaining files (or publications with files) that have unknown versions.

Does that make sense to everyone, or do I have blindspots?

Smullz622 (dismissed):
lgtm!

app/models/publication.rb
```diff
@@ -154,6 +154,13 @@ def self.preferred_versions
             .where(%{oa_workflow_state IS DISTINCT FROM 'oa metadata search pending'})
             .where(%{oa_status_last_checked_at IS NULL OR oa_status_last_checked_at < ?}, 1.hour.ago)
         }
+  scope :file_version_check_failed, -> {
+    activity_insight_oa_publication
+      .where.not(publications: { preferred_version: nil })
```
EricDurante, line 159: I think this line can just be
```ruby
.where.not(preferred_version: nil)
```

PR #720, opened 2023-03-02
Grab permissions regardless of stored file versions

ajkiessl:
Grabbing permissions per publication rather than per file.  The permissions are pulled and stored with the publication regardless of the versions of the files we have stored.

### DIFF
--- a/app/services/oa_workflow_service.rb
+++ b/app/services/oa_workflow_service.rb
@@ -17,12 +17,12 @@ class OAWorkflowService
       FetchOAMetadataJob.perform_later(pub.id)
     end
 
-    ActivityInsightOAFile.needs_permissions_check.each do |file|
-      file.version_checked = true
-      file.save!
-      PermissionsCheckJob.perform_later(file.id)
+    Publication.needs_permissions_check.each do |pub|
+      pub.permissions_last_checked_at = Time.current
+      pub.save!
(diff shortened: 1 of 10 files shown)


app/services/oa_workflow_service.rb
```diff
@@ -17,12 +17,12 @@ def workflow
...
-      file.version_checked = true
-      file.save!
-      PermissionsCheckJob.perform_later(file.id)
+    Publication.needs_permissions_check.each do |pub|
+      pub.permissions_last_checked_at = Time.current
+      pub.save!
+      PermissionsCheckJob.perform_later(pub.id)
     rescue StandardError
-      file.update_column(:version_checked, true)
+      pub.update_column(:permissions_last_checked_at, Time.current)
```
ajkiessl, line 25: I'm using the `.permissions_last_checked_at` timestamp in place of the `version_checked` boolean.  It functions basically the same, but it's kind of ad hoc.  Not sure if a boolean would be better.

Smullz622: I think `.permissions_last_checked_at` was originally added for this function because we thought we might be checking permissions more than once, but then we changed directions. I'm not sure if a boolean would be better, but if that is what we go with, I don't think we need `.permissions_last_checked_at` anymore.

EricDurante (approved):
This looks good to me.

PR #722, opened 2023-03-03
adds download files job

### LINKED ISSUE #691: Download publication files from Activity Insight
After a publication that is moving through the Activity Insight open access workflow has had its DOI confirmed and its open access status checked, the next step will be check to see what version of the publication was uploaded. In order to do this, we'll have to actually download and store a local copy of the file that was uploaded to Activity Insight by the faculty member so that we can analyze the file with our existing (and currently in development) version checking tools (https://github.com/psu-libraries/researcher-metadata/issues/575), and so that an RMD admin can view the file if the automated version checking is inconclusive.

Like the other steps in the workflow, this step will begin with the overall workflow cron job queuing up a background job to handle the file download. The cron job should look for publications that have an Activity Insight file upload location, a confirmed DOI, and open access metadata that indicates that the publication should be deposited in Scholarsphere if possible. At this point, we'll be looking for publications that have successfully gone through the automated open access metadata check from the previous step and still do not have any associated open access locations. We'll also need to filter out publications with files that have already been downloaded and publications that already have pending download job on the queue.

Once a file has been downloaded, we may ultimately want to do the version checking  as part of the same step (background job), or we may find that it makes more sense for the version checking process to be handled by a separate job, but for the purposes of this issue, we'll consider it complete whenever the files are being downloaded.

As for the mechanics of actually downloading the files, we may be able to incorporate code from the [fams_tools](https://github.com/psu-libraries/fams_tools) repository (which has similar functionality) with the ultimate goal being to deprecate and eventually discontinue use of fams_tools for this purpose.

Smullz622:
Fixes #691

### DIFF
--- a/spec/component/jobs/publication_download_job_spec.rb
+++ b/spec/component/jobs/publication_download_job_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'component/component_spec_helper'
+
+describe PublicationDownloadJob, type: :job do
+  let(:job) { described_class }
+
+  describe '.perform_later' do
+    ActiveJob::Base.queue_adapter = :test
+    it 'enqueues a job' do
+      expect { job.perform_later(1) }.to have_enqueued_job.with(1).on_queue('default')
+    end
+  end
+
+  describe '#perform_now' do
+    let!(:publication) { create(:publication)}
+    let!(:ai_oa_file) { create(:activity_insight_oa_file, publication: publication, version: 'acceptedVersion') }
+    let(:returned_file) { fixture_file_upload('test_file.pdf', 'application/pdf') }
+
+    before do
+      allow(HTTParty).to receive(:get).with().and_return(returned_file) #add AI request
+    end
+
+    it 'saves the new file path' do
+      job.perform_now(ai_oa_file.id)
+      expect(ai_oa_file.stored_file_path).to eq "tmp/uploads/cache/activity_insight_file_uploads/#{model&.id || object_id}/file"
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


spec/component/jobs/publication_download_job_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    it 'saves the new file path' do
+      job.perform_now(ai_oa_file.id)
+      expect(ai_oa_file.reload.stored_file_path).to eq Rails.root.join("tmp/uploads/activity_insight_file_uploads/#{ai_oa_file.id}/file/location").to_s
+    end
+
+    it 'uploads the file' do
+      job.perform_now(ai_oa_file.id)
+      expect(File.open(Rails.root.join("tmp/uploads/activity_insight_file_uploads/#{ai_oa_file.id}/file/location")).size).to eq 13103
+    end
+  end
```
ajkiessl, line 34: Does this block of tests actually store a file locally?  If so, should we delete it after these tests complete?

ajkiessl:
Hmm not sure why SimpleCov isn't working.  Maybe the tests just need ran again?  

Other than my one question above this LGTM.

spec/component/jobs/publication_download_job_spec.rb
EricDurante: @ajkiessl This has changed, but the old test actually did download a file and the new test that has replaced this one also downloads and stores an actual file. The test file is very small, and I think the same file gets overwritten on each test run, so I'm pretty sure that it won't just pile up new files indefinitely. We could delete the file after the test runs, but I don't feel too strongly about it either way. Having the file remain after the test was finished was helpful for debugging some issues.

ajkiessl: Ok, sounds good.  I was concerned about the files piling up, but if that won't happen it's not a problem.

ajkiessl:
The glacial test fixes were merged to main so those changes can now be merged here.

ajkiessl (approved):
LGTM Thanks!

PR #725, opened 2023-03-06
Editing activity insight oa files

### LINKED ISSUE #689: Manual editing of file versions
We do not have any edit capabilities for `ActivityInsightOaFile` within rails_admin right now.  We'll have to add that.  Then, we can link to the `ActivityInsightOaFile` edit page from the manual file version checking list.

Admins will likely want to download and view the file at this point.  We should already have the file locally, so we should link to that for the admin to view. ~~We can either directly connect to S3 in RMD, or we can integrate with the https://github.com/psu-libraries/ai-s3-authorizer .  I'm leaning toward setting up a direct S3 connection in RMD.  It may be easier to maintain that way, and be more reliable.~~

We'll want a link in the edit page for admins to download the file.  We could also make the file location in the bucket list a link to download the file.

ajkiessl:
Adds a link from the file version check failed list to the edit page for activity insight oa files:

Screenshot 2023-03-06 at 3 12 32 PM

Adds the rails admin show, list, and edit pages.  Here's the edit page:

Screenshot 2023-03-06 at 3 12 53 PM

Also, adds a validation that restricts the version to nil, unknown, acceptedVersion, or publishedVersion.  This does not include any links to download/view the file.  I'm going to wait for #722 before doing that.

closes #689


EricDurante (approved):
Looks good

PR #726, opened 2023-03-07
Detect new files when importing Activity Insight OA Files

### LINKED ISSUE #693: Detect changes to existing file uploads in Activity Insight and process new files
A big problem with the existing manual workflow for depositing publications from Activity Insight is that a faculty member can go back into Activity Insight at any time, edit one of their publications, and replace an old file upload with a new one that may be a different version of the publication. When this happens, the librarians who are working on gathering information in order to try to deposit the older file upload may not realize that a new file has been uploaded until after they've already spent time working on the version that they have - and sometimes it turns out that the older version that they've been working on cannot be deposited.

We'd like to try to save some wasted effort by automatically detecting when an existing publication with an existing file upload from Activity Insight has a new file upload in Activity Insight. We can skip this process altogether for publications that we know to be open access already and for publications for which we already have a file version that can be deposited. For all other publications with an associated file upload, we should check periodically to see if the file (or file location) that we have is still the same file that is stored in Activity Insight. If it's not, then we should import the location metadata for this new file (while preserving the metadata for the older file). The cron job that orchestrates this workflow should automatically detect publications that have new file metadata that is missing version or permissions information, and it should queue up jobs to download the new file and fetch this information.

ajkiessl:
Adds criteria for when _not_ to import an Activity Insight OA File:

- When an existing file exists and it is the correct/valid version
- When the `location` of the file being imported is the same as the `location` of the existing file  

These changes will make it so the OA Workflow can "reset" (or atleast reset back to the file version checking stage) if someone originally uploaded the wrong version, deletes this file in Activity Insight, and uploads a new file that is hopefully the correct version.

closes #693

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -203,13 +203,10 @@ class ActivityInsightImporter
 
           activity_insight_file_location = pub.postprints&.first&.location
 
-          if pub_record.no_open_access_information? && activity_insight_file_location.present? && Publication.oa_publication_types.include?(pub_record.publication_type)
-            aif = pub_record.activity_insight_oa_files.first
+          if activity_insight_file_location.present? && pub_record.can_receive_new_ai_oa_files?
+            aif = pub_record.activity_insight_oa_files.find_by(location: activity_insight_file_location)
 
-            if aif.present?
-              aif.update location: activity_insight_file_location
-              aif.save!
-            else
+            unless aif.present?
               file = ActivityInsightOAFile.create(location: activity_insight_file_location)
               pub_record.activity_insight_oa_files << file
               pub_record.save!
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -210,7 +210,11 @@ class ActivityInsightImporter
               file = ActivityInsightOAFile.create(location: activity_insight_file_location)
               pub_record.activity_insight_oa_files << file
               pub_record.save!
-              DOIVerificationJob.perform_later(pub_record.id)
+              unless pub_record.doi_verified == true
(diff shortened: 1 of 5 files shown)


app/importers/activity_insight_importer.rb
```diff
@@ -203,13 +203,10 @@ def call
...
+          if activity_insight_file_location.present? && pub_record.can_receive_new_ai_oa_files?
+            aif = pub_record.activity_insight_oa_files.find_by(location: activity_insight_file_location)
 
-            if aif.present?
-              aif.update location: activity_insight_file_location
-              aif.save!
-            else
+            if aif.blank?
               file = ActivityInsightOAFile.create(location: activity_insight_file_location)
               pub_record.activity_insight_oa_files << file
```
EricDurante, line 211: I think that we need to do [this](https://github.com/psu-libraries/researcher-metadata/blob/47cd1faaa854d865da95fe90480b1a20bded8225/app/services/oa_workflow_service.rb#L6) here also if we're going to queue up the DOI verification job at this point.

EricDurante (changes requested):
Looks good to me apart from my one comment that's actually unrelated to your current changes.

app/importers/activity_insight_importer.rb
ajkiessl: Ah good catch.  I'll fix that up.

app/importers/activity_insight_importer.rb
```diff
@@ -203,17 +203,18 @@ def call
...
               pub_record.save!
-              DOIVerificationJob.perform_later(pub_record.id)
+              unless pub_record.doi_verified == true
```
ajkiessl, line 213: @EricDurante I also added some logic to not do the doi verification if the doi is already verified.

EricDurante (approved):

PR #729, opened 2023-03-08
Condense shared integration spec examples for better test performance

EricDurante:


ajkiessl (approved):

EricDurante:
@ajkiessl @whereismyjetpack 

There's still more refactoring like this that can be done, but this has the test suite running in just under 8 minutes on my machine with exactly the same test coverage.

Go ahead and review/merge this if you want so that we can benefit from what's already done. If I get a chance to do any more, I'll open another PR.

ajkiessl:
Looks good, thanks!

PR #730, opened 2023-03-08
Fix failing glacial test

ajkiessl:
Now parses orcid response as json, and does a better job off asserting the response is the most recently sent record.

### DIFF
--- a/spec/component/models/orcid_integration_spec.rb
+++ b/spec/component/models/orcid_integration_spec.rb
@@ -39,9 +39,13 @@ describe OrcidAPIClient do
     }
     let(:employments_uri) { "#{described_class.base_uri}/#{resource.orcid_id}/employments" }
     let(:get_employments) { client.class.get(employments_uri, headers).parsed_response }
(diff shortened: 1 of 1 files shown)


EricDurante (changes requested):
Looks good - thanks!

If this only takes 4 seconds to run, then I say we just remove the `glacial` tag and run it all of the time.

ajkiessl:
Since this test creates a new record in the orcid sandbox every time it runs, I think we may have made this glacial to reduce the number of records we create.   I don't know if there's a limit to the number of records we can add, but I'm fine with removing the glacial tag and seeing what happens.

ajkiessl:
Actually, it's supposed to delete the record.  🤔 Maybe I need to take another look at this.

ajkiessl:
Ok, this should be good now.  It deletes the record after creating it.  The test takes a little longer now, but not too long (~6 seconds).

EricDurante (approved):
👍

PR #781, opened 2023-08-17
Fix ai download

ajkiessl:
The escaping seemed to work for me.  There were 344 `ActivityInsightOAFiles` ready for download on my local machine.  After running this with the escaping implemented, 343 successfully downloaded and one failed due to a timeout error.  I added some error handling to deal with these.  I went with rescuing `StandardError` since that seems to be the preferred way to cover all of the potential errors with Net::HTTP.


EricDurante (approved):
Yep, this fixed the issue for me.

PR #786, opened 2023-08-21
adds activity insight oa file links

### LINKED ISSUE #728: Add links to download/view Activity Insight Oa File
The location on the edit and show pages for the ActivityInsightOAFiles should be links to download the file.

It'd probably be helpful to do this in the index list as well, so let's to that.

Also, somewhere in the permissions review list/s in the OA Workflow we'll want links to download/view the files.  This may need to be a separate link or button from the locations since they'll probably be used to link to the edit pages.

Smullz622:
Fixes #728

### DIFF
--- a/app/controllers/activity_insight_oa_workflow/files_controller.rb
+++ b/app/controllers/activity_insight_oa_workflow/files_controller.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class ActivityInsightOAWorkflow::FilesController < ActivityInsightOAWorkflowController
+  def download
+    file = ActivityInsightOAFile.find(params[:activity_insight_oa_file_id])
+    send_file(file.stored_file_path)
+  end
+end
--- a/app/models/activity_insight_oa_file.rb
+++ b/app/models/activity_insight_oa_file.rb
@@ -54,6 +54,11 @@ class ActivityInsightOAFile < ApplicationRecord
       field(:created_at)
       field(:updated_at)
       field(:publication)
+      field 'File' do
+        formatted_value do
+          bindings[:view].link_to("Download #{bindings[:object].download_filename}", Rails.application.routes.url_helpers.activity_insight_oa_workflow_file_download_path(bindings[:object].id))
+        end
+      end
     end
 
     list do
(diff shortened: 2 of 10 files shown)


ajkiessl (changes requested):
I think we'll need to remove this line or else we won't have files downloaded at the DOI Verification and Permissions Review stages:
https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/activity_insight_oa_file.rb#L12
This will probably break some tests.

The right edge of the table is getting funky with long filenames in the DOI Verification list and the Permission Review page too.  The Version Review page looks good, though, so I think there might just be some HTML/CSS reworking that could fix this.
Screenshot 2023-08-21 at 1 38 16 PM

Also, I think we should remove "Download" in the text in the links to download the files.  It kind of looks like "Download" is a part of the filename.  Maybe we could rename the columns and labels to "File Download" instead of just "File".

EricDurante:
> I think we'll need to remove this line or else there we won't have files downloaded at the DOI Verification and Permissions Review stages:
https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/activity_insight_oa_file.rb#L12

It's probably my fault that this was added in the first place. I likely assumed that we shouldn't download the file until after we passed the permission check (and knew for sure that the publication could be deposited), but I didn't realize that the file may be useful during the permissions check itself and the manual review steps that come before.

spec/component/services/oa_workflow_service_spec.rb
```diff
@@ -111,7 +111,7 @@
       it 'calls the publication download job with that file' do
         service.workflow
         expect(PublicationDownloadJob).to have_received(:perform_later).with(activity_insight_oa_file4.id)
-        expect(PublicationDownloadJob).not_to have_received(:perform_later).with(activity_insight_oa_file5.id)
+        expect(PublicationDownloadJob).to have_received(:perform_later).with(activity_insight_oa_file5.id)
```
ajkiessl, line 114: I think we should test at least one of the other files that isn't sent through the job.  Maybe `file6`?

ajkiessl (changes requested):
Just one minor, thing.  Other than that, it looks good!

ajkiessl (approved):
LGTM

PR #788, opened 2023-08-21
780 permissions metadata

EricDurante:


ajkiessl (approved):
This all looks good to me.  I didn't know you could do the grouping thing with rails admin.  That makes the show and edit pages much easier to navigate 👍

Issue #692, opened 2023-02-09
Retry workflow for files that cannot be deposited


EricDurante:
This issue is a placeholder for the whole branch of the workflow that will occur after we've identified the versions and permissions for all of a publication's files and determined that we don't have any files that can be deposited (see https://github.com/psu-libraries/researcher-metadata/issues/690). It may end up needing to be decomposed into a handful of separate issues depending on how we want this workflow to behave.

Some things that we may want to do automatically when we encounter this condition:
- delete the files that can't be deposited from Activity Insight (if possible)?
- update the publication's status in Activity Insight?
- delete the local copies of files that cannot be deposited from RMD and update RMD's associated database records to indicate that the files were not able to be deposited
- keep any metadata that we've gathered earlier in the workflow (confirmed DOI, open access information, etc.) that will be useful in the next iteration of the workflow if/when we receive a new file for the publication
- either notify the faculty member directly that we couldn't deposit the file that they provided, or notify @nmg110 so that she can contact the faculty member and help them to provide the correct version

nmg110:
RE: either notify the faculty member directly that we couldn't deposit the file that they provided, or notify @nmg110 so that she can contact the faculty member and help them to provide the correct version

If it is a file that we cannot deposit what if they then get thrown in the OA Email group that automatically goes out from RMD? And then a list is of accounts, AI ID's, Titles, and file names are emailed to AI-Support@psu.edu for those attachments to be removed from AI? Just a thought.

ajkiessl:
Sometimes faculty is not sure what the versions are and sending an email directly to them confuses them when the version is wrong, so sending an email to the OA group might be a better option than send to the faculty.

ajkiessl:
We'll need to determine what exactly kicks this off.

EricDurante:
Thanks for uncovering this additional complexity right away, @anaelizabethenriquez.

I supposed that it's possible that if one author has a batch of publications that require action, then each of those publications may require action for a different one of the reasons that you listed?

I think that by the time that a publication/file reaches this point in the workflow, we'd have enough information about the version to automatically specify the reason why the file was rejected, so that may not be an issue. If that turns out to not be the case, then we could allow the RMD admin to choose a reason (or even a whole different email template) for rejecting the file.

I also don't think that it would be too difficult to group files in this list by the Activity Insight users that uploaded them and then send one email per group that lists each file. If the files were grouped in this way, would you still want the option to send an email per file in addition to the batch email?

[...]

Issue #789, opened 2023-08-22
Allow both Published and Accepted version as `preferred_version`


ajkiessl:
This will require adding a third potential `preferred_version`.  We can call it "Published or Accepted".  The permissions data import will have to be modified so that if a publication can have both, this value is set.  We'll also have to alter the logic for what constitutes a "failed" version check/match.

anaelizabethenriquez:
Thinking about this more, I think I got confused before. Sorry about that. There may be different values for License, Set Statement, and Embargo depending on whether it is the Published Version or the Accepted Version being deposited. So are we sending everything like this for manual permissions review once we figure out which version the file from Activity Insight is, and expecting an admin to update the preferred version to whatever we've got and fill in permissions metadata as needed? @ajkiessl

EricDurante:
@anaelizabethenriquez Hmm, well the way that we have things working now, license, set statement, and embargo date are properties of the publication in RMD rather than properties of each of the files, and they would be the values for whatever the "preferred version" (also a property of the publication) is.

In the case where either version can be deposited, I guess that there are a few ways that it could be handled. I think that the least nice way would be to leave things as they are and do what you described - expect the admins to manually check and possibly correct the permissions metadata for whichever version is actually being deposited.

A nicer way might be to tweak our data model and do some additional metadata importing. One way to do this might be to make license, set statement, and embargo date properties of the files and then also change the workflow to automatically collect this metadata for each file _after_ we know the file version. However, I think that we could also do something similar without changing the existing workflow. We could model permissions metadata in the same way that OA Button does. In other words, a publication in RMD would could have multiple sets of permissions (a "permission" would be its own data model with the license, etc. properties), and one or more of those could be marked as "preferred" or "acceptable". We could gather all of this metadata from OA Button during the existing automated permissions check step and then use it during the final steps to determine if we have an acceptable version. If we do, then we'd already have the right metadata for that version, and we could automatically use that set of permissions for the chosen file. I don't have a strong feeling about which of these two approaches would be better yet..

ajkiessl:
Yeah, the second option sounds better, but would likely be more time consuming.  I think it could get done in the next two sprints, though.  I don't think the first option would require any _big_ changes, and could hold us off until we get to another round of development.

I'm inclined to do the second option just so things are more thoroughly done from the start, but if there are other high priority things we want to get done, and the first option suffices in the meantime, then it might be better to do option 1.

EricDurante:
[...]

PR #800, opened 2023-08-25
Adds wrong file version list

### LINKED ISSUE #719: List for Publications whose file versions are all determined but the wrong version
We'll want to keep these in their own list separate from #688.  At this point in the workflow there's the chance the author will need to re-upload a file with the proper version.  We haven't entirely figured out how this will happen and what automation we can do with it (#692), but we do know that we want the admins to have a list to view these publications and their files.  The list should probably include the publication's title, its preferred version, the file locations (with a link to view the file or edit the file record), and the version of each file.  We'll also want a link to edit the publication in case an admin wants to view other metadata for the publication or change some metadata.

Smullz622:
Fixes #719 and #692 

I still need to add tests for everything, but I wanted to get this up for initial review while I'm away.


app/mailers/faculty_notifications_mailer.rb
```diff
@@ -19,4 +19,13 @@ def scholarsphere_deposit_failure(user, deposit)
          from: 'scholarsphere@psu.edu',
          reply_to: 'scholarsphere@psu.edu'
   end
+
+  def wrong_file_version(publications)
+    @publications = publications
+    #will need to change the mail to to whatever connects a publication to the activity insight uploader's email
```
EricDurante, line 25: This is now an attribute of the file record. So `ActivityInsightOAFile#user` is the RMD user who uploaded the file in Activity Insight. This change has been merged into main.

app/views/activity_insight_oa_workflow/wrong_file_version_curation/index.html.erb
```diff
@@ -0,0 +1,69 @@
+<div class="container pt-4 pb-5">
+  <%= link_to '<< Back', activity_insight_oa_workflow_path %>
+  <h1 class="pb-3 pt-2">Publications with Wrong File Version</h1>
+  <div class='card'>
```
EricDurante, line 4: I removed this `.card` container in all of the other list views to allow the larger tables to take up the full width of the screen. We'll probably want to remove it here as well.

app/models/publication.rb
```diff
@@ -159,6 +159,13 @@ def self.preferred_versions
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND publications.preferred_version = activity_insight_oa_files.version)})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version IS NULL)})
   }
+  scope :wrong_file_version, -> {
+    activity_insight_oa_publication
+      .where.not(preferred_version: nil)
+      .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version = 'unknown')})
+      .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND publications.preferred_version = activity_insight_oa_files.version)})
```
EricDurante, line 166: Just a heads-up that this is going to get more complicated when we get to https://github.com/psu-libraries/researcher-metadata/issues/789, but we'll cross that bridge when we come to it.

ajkiessl (changes requested):
I have one UI change request below.  Also, in that change request, I mention something to consider about how we select the author.  I don't think we need to do anything about that, though.  I still need to more thoroughly look at the tests, but I just wanted to get this one request out.

app/views/activity_insight_oa_workflow/wrong_file_version_curation/index.html.erb
```diff
@@ -0,0 +1,65 @@
...
+          <td rowspan="<%= p.activity_insight_oa_files.count %>">
+            <% if publications.index(p) == 0 %>
+              <%= User.find(p.activity_insight_oa_files.first.user_id).webaccess_id %>
+            <% end %>
+          </td>
+          <td rowspan="<%= p.activity_insight_oa_files.count %>">
+            <% if publications.index(p) == 0 %>
+                <%= button_to "Send Batch Email", activity_insight_oa_workflow_wrong_file_version_email_path, params: {publications: publications} %>
+            <% end %>
+          </td>
```
ajkiessl, line 32: We move some things around here and get the table nesting a little more aligned:

```suggestion
      <tr id="<%= dom_id(User.find(author)) %>">
      <td rowspan="<%= publications.collect(&:activity_insight_oa_files).flatten.count %>">
        <%= User.find(author).webaccess_id %>
      </td>
      <td rowspan="<%= publications.collect(&:activity_insight_oa_files).flatten.count %>">
       <%= button_to "Send Batch Email", activity_insight_oa_workflow_wrong_file_version_email_path, params: {publications: publications} %>
      </td>
      <% publications.each do |p| %>
```

This gives us something like this:

Screenshot 2023-09-06 at 3 25 18 PM

Also, there's the chance we end up with multiple files for a single publication where the authors are different.  Since we select the first OA file for every publication to base the author off of, and also do that in the email sending, I think we should be fine.

ajkiessl (changes requested):
I have one UI change request below. Also, in that change request, I mention something to consider about how we select the author. I don't think we need to do anything about that, though. I still need to more thoroughly look at the tests, but I just wanted to get this one request out.

app/controllers/activity_insight_oa_workflow/wrong_file_version_curation_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ActivityInsightOAWorkflow::WrongFileVersionCurationController < ActivityInsightOAWorkflowController
+  def index
+    @publications = Publication.wrong_file_version.order('wrong_oa_version_notification_sent_at DESC NULLS FIRST')
+  end
+
+  def email_author(publications)
```
EricDurante, line 8: It's not clear to me how a collection of publications is getting passed in here. Our code doesn't call the controller method directly, the router does, and the router doesn't pass arguments to the method. I don't see an end-to-end integration test for this, so if I had to guess, I'd bet that clicking an email button is sending the email without any publication data.

However, it looks like the form (button) for sending the email is being constructed such that it's posting the publications as form data. That's what we want, but that data won't be passed into the controller method as an argument (and definitely not as fully-formed ActiveRecord objects). Instead, I think we'll need to pull the publication IDs out of the params hash and then find the records. Something like this:

```ruby
publications = Publication.wrong_file_version.find(params[:publications])
```

I'm also not totally sure whether or not the way in which the publications are being passed to `button_to` as params will cause the collection of publication objects to be transformed into an array of publication IDs, but that's what we'd need.

spec/integration/admin/activity_insight_oa_workflow/wrong_file_version_review_spec.rb
EricDurante, line 1: We should definitely have an integration test that exercises the email-sending functionality through the UI and verifies that the emails are being sent with the correct content.

EricDurante (changes requested):
@Smullz622 It doesn't look to me like this functionality is completely working yet. Let me know if you have any questions about my comments and/or if you'd like to pair up to take a closer look at any of this. I'd be happy to help!

app/mailers/faculty_notifications_mailer.rb
```diff
@@ -19,4 +19,12 @@ def scholarsphere_deposit_failure(user, deposit)
          from: 'scholarsphere@psu.edu',
          reply_to: 'scholarsphere@psu.edu'
   end
+
+  def wrong_file_version(publications)
+    @publications = publications
+    mail to: "#{User.find(@publications.first.activity_insight_oa_files.first.user_id).webaccess_id}@psu.edu",
```
EricDurante, line 25: I think that this line could be refactored by using the new Publication#activity_insight_file_uploader method.

It could possibly be refactored further by wrapping the user object in an instance of the UserProfile class which already has a method for assembling the email address for a user.



Something like
```ruby
profile = UserProfile.new(@publications.first.activity_insight_file_uploader)
mail to: profile.email
```

That's not really any more concise, but it eliminates a bunch of duplication and method chaining, and it improves readability - all of which is far more important.

spec/requests/activity_insight_oa_workflow/wrong_file_version_curation/email_author_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'support/authentication'
+
+describe ActivityInsightOAWorkflow::WrongFileVersionCurationController do
+  describe 'POST /activity_insight_oa_workflow/wrong_file_version_curation/email_author' do
+    let(:user) { create(:user, is_admin: true) }
+
+    before do
+      sign_in_as(user)
+      post user_azure_oauth_omniauth_authorize_path
+    end
```
EricDurante, line 13: We can leave this for another time, but just a note here to say that we might need to handle authentication in same way that we are in the existing controller specs - just by totally mocking out the entire authentication mechanism down at the rack middleware level.

i.e. https://github.com/psu-libraries/researcher-metadata/blob/fac721ad2a7452a17fd0d0421f59693fb96c5dfb/spec/component/controllers/activity_insight_oa_workflow_controller_spec.rb#L16

app/models/publication.rb
```diff
@@ -216,6 +223,10 @@ def confirmed_users
     users.where(authorships: { confirmed: true })
   end
 
+  def activity_insight_file_uploader
```
EricDurante, line 226: Sorry, I have second thoughts on the name of this method - but only because there's a class in this application named `ActivityInsightFileUploader`, which is not at all what this method returns. I'd be less confused if it was called something like `activity_insight_upload_user`.

EricDurante (approved):
Suggested a couple of minor refactorings that I think are worth doing, but I'll leave to your discretion.

Looks good to me otherwise.

spec/requests/activity_insight_oa_workflow/wrong_file_version_curation/email_author_spec.rb
ajkiessl: FWIW I was able to get this to work by adding:
```
RSpec.configure do |config|
  config.include Devise::Test::IntegrationHelpers, type: :request
end
```
to the `requests_spec_helper.rb` and then just using `sign_in(user)` rather than:
```
 before do
      sign_in_as(user)
      post user_azure_oauth_omniauth_authorize_path
 end
```

app/controllers/activity_insight_oa_workflow/wrong_file_version_curation_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def email_author
+    publications = Publication.wrong_file_version.find(params[:publications])
+
+    FacultyNotificationsMailer.wrong_file_version(publications).deliver_now
+    ActiveRecord::Base.transaction do
+      publications.each do |pub|
+        pub.update_column(:wrong_oa_version_notification_sent_at, Time.current)
+      end
+    end
+    flash[:alert] = "Email sent to #{publications.first.activity_insight_file_uploader.webaccess_id}}"
```
EricDurante, line 17: oops - one other thing. And this is my fault (I told you the wrong thing). This should be `flash[:notice]` instead of `flash[:alert]`. We're using the former for success messages and the latter for error messages, and I can never remember which one is which off of the top of my head.

spec/requests/activity_insight_oa_workflow/wrong_file_version_curation/email_author_spec.rb
EricDurante: Oh, awesome! Thanks, @ajkiessl!

`sign_in(user)` looked like it should have worked - I just didn't realize that I was missing the necessary configuration.

EricDurante (approved):
Looks good!

PR #801, opened 2023-08-28
User file association

EricDurante:
@ajkiessl

This PR contains the new association between `User` and `ActivityInsightOAFile` records. It also contains the new list of publications that are ready for metadata review prior to deposit. The list functionality isn't complete yet, but I wanted to get this merged in at this point since it's all working and several other things will depend on the user/file association.

There's also a temporary change to the Activity Insight import that will populate the `User` association for existing `ActivityInsightOAFile` records. We don't need to release this to production right away if you don't want, but after it has eventually been released and the updated Activity Insight import has been run, we can undo that change.

### DIFF
--- a/app/models/activity_insight_oa_file.rb
+++ b/app/models/activity_insight_oa_file.rb
@@ -3,6 +3,7 @@
 class ActivityInsightOAFile < ApplicationRecord
   belongs_to :publication,
              inverse_of: :activity_insight_oa_files
+  belongs_to :user
 
   mount_uploader :file_download_location, ActivityInsightFileUploader
 
--- a/app/models/activity_insight_oa_file.rb
+++ b/app/models/activity_insight_oa_file.rb
@@ -63,7 +63,7 @@ class ActivityInsightOAFile < ApplicationRecord
         end
       end
       field(:downloaded)
-      field(:download_location_value) { label 'File download'}
+      field(:download_location_value) { label 'File download' }
     end
 
     list do
@@ -74,7 +74,7 @@ class ActivityInsightOAFile < ApplicationRecord
       field(:updated_at)
       field(:publication)
       field(:downloaded)
-      field(:download_location_value) { label 'File download'}
+      field(:download_location_value) { label 'File download' }
     end
 
     edit do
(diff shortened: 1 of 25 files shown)


app/models/activity_insight_oa_file.rb
```diff
@@ -58,6 +69,8 @@ def version_status_display
           bindings[:view].link_to("Download #{bindings[:object].download_filename}", Rails.application.routes.url_helpers.activity_insight_oa_workflow_file_download_path(bindings[:object].id))
         end
       end
+      field(:downloaded)
+      field(:download_location_value) { label 'File download' }
```
ajkiessl, line 73: Could we consolidate the 'File' and 'File download' field?  Maybe just have a single 'File download' field with this in it:

`bindings[:view].link_to("#{bindings[:object].download_location_value}", Rails.application.routes.url_helpers.activity_insight_oa_workflow_file_download_path(bindings[:object].id))`

app/views/activity_insight_oa_workflow/permissions_curation/index.html.erb
```diff
@@ -1,7 +1,6 @@
 <div class="container pt-4 pb-5">
   <%= link_to '<< Back', activity_insight_oa_workflow_path %>
   <h1 class="pb-3 pt-2">Publications Requiring Permissions Review</h1>
-  <div class='card'>
   <table class="table table-scrollable">
```
ajkiessl, line 4: If we remove the card container, we should remove table-scrollable class.  It will probably be more useful for admins to be able to see the whole table without scrolling.

ajkiessl (changes requested):
This all looks good except for a few minor things.  I'll add an issue to remove the temporary user linking import things.  I think we can plan on releasing this PR with the next release two weeks from now.  Then, work on removing the user linking code in the next sprint.

app/views/activity_insight_oa_workflow/permissions_curation/index.html.erb
EricDurante: Yeah, my thought was that if the tables are getting wide, then better to allow them to grow to the size of the screen rather than hiding part of them inside of a container with a fixed size. If they get so wide that you have to scroll the whole screen, then maybe that's another issue, but we're not there yet...

app/models/activity_insight_oa_file.rb
EricDurante: Yeah, good call.

ajkiessl (approved):
🚀

PR #804, opened 2023-08-29
adds rake tasks for pulling rmd data from s3

whereismyjetpack:
Adds two rake tasks to facilitate getting a database dump into a local environment. These rake tasks assume that your local aws environment is setup. e.g (`eval $(aws configure --profile=default export-credentials --format env)`)

to pull from another bucket, you can set the `AWS_BUCKET` and `AWS_PREFIX` parameters. they are defaulted to the rmd backup bucket.

`database_data:download_from_s3` will look for the files in the bucket prefix, allow a user to select one, and download it to `Rails.root/tmp`

`database_data:load_from_backup` when called without a dump_file parameter will call download_from_s3, and then load that file into the development database. 

`database_data:load_from_backup:[/path/to/dumpfile']` will skip downloading from s3 again, and load in the passed file.

### DIFF
--- a/lib/tasks/database_data.rake
+++ b/lib/tasks/database_data.rake
@@ -34,7 +34,7 @@ namespace :database_data do
     download_from_s3
   end
 
-  task :load_from_backup, [:dump_file] => :environment do | _t, args|
+  task :load_from_backup, [:dump_file] => :environment do |_t, args|
     desc 'Loads a database backup into the development environment'
     filename = args[:dump_file]
     db_config = Rails.configuration.database_configuration
--- a/lib/tasks/database_data.rake
+++ b/lib/tasks/database_data.rake
@@ -18,15 +18,15 @@ namespace :database_data do
     client = Aws::S3::Client.new
 
     download_prog = ProgressBar.create(total: nil, title: 'download', format: '%t |%B| %a')
-    objects = client.list_objects({bucket: bucket, prefix: prefix}).contents.reverse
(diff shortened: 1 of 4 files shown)


lib/tasks/database_data.rake
```diff
@@ -9,6 +11,64 @@ namespace :database_data do
...
+    psql_prog = ProgressBar.create(total: nil, title: 'restore', format: '%t |%B| %a')
+    Thread.new { until psql_prog.finished? do psql_prog.increment; sleep(0.2); end }
+    `pg_restore --no-owner --no-acl --dbname #{db_config['development']['database']} #{filename}`
+    psql_prog.finish
+
+    # Catch up on any migrations
+    ActiveRecord::Tasks::DatabaseTasks.migrate
+  end
+
   task get_prod_data_file: :environment do
```
ajkiessl, line 72: We can delete the tasks below now.  They don't work anymore; not since the server auth changes (or whatever happened).

lib/tasks/database_data.rake
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require 'progressbar'
+require 'aws-sdk-s3'
+require 'tty-prompt'
 
 namespace :database_data do
   # These tasks are for getting production data into local, qa, and stage environments
```
ajkiessl, line 8: We should probably delete these comments too.

ajkiessl (changes requested):
Just some things to remove below 👇

ajkiessl (approved):
LGTM

PR #805, opened 2023-08-29
Cleanup secondary titles

ajkiessl:
Removes rake task that no longer works.  Adds SubtitleCleanupService to remove secondary titles from pubs if it is contained in the title.  Rake task and test for this.  Removes secondary title check in DoiVerificationService

### DIFF
--- a/app/services/doi_verification_service.rb
+++ b/app/services/doi_verification_service.rb
@@ -10,7 +10,7 @@ class DOIVerificationService
   end
 
   def verify
-    pub_title = publication.matchable_title + publication.matchable_secondary_title
+    pub_title = publication.matchable_title
     unpaywall_title = UnpaywallClient.query_unpaywall(publication).matchable_title
     publication.doi_verified = compare_title(pub_title, unpaywall_title)
     publication.save!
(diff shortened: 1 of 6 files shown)


ajkiessl:
Niftany and tests are passing locally, at least.  Still working on getting CI working.

EricDurante:
Does anything about [this](https://github.com/psu-libraries/researcher-metadata/blob/f6cdc0c38d797b608dd5b41c18e1934402aad643/app/models/publication_merge_on_matching_policy.rb#L64) need to change? I have no idea - just asking to make sure that nothing gets accidentally overlooked.

ajkiessl:
> Does anything about [this](https://github.com/psu-libraries/researcher-metadata/blob/f6cdc0c38d797b608dd5b41c18e1934402aad643/app/models/publication_merge_on_matching_policy.rb#L64) need to change? I have no idea - just asking to make sure that nothing gets accidentally overlooked.

Good catch.  Yeah, this probably should be changed sooner rather than later.  I will add it to this PR.  Since the Pure imports are fixed to put the subtitle in the title, and the script in this PR will clean up the remaining issues with subtitles, we don't really need this anymore.  In fact, it could cause issues given how much deduplication is happening now.

ajkiessl:
@EricDurante This is done and ready for review whenever you get a chance.

EricDurante (approved):
I'm not close enough to the details about the data here to judge the correctness of the behavior, but the code looks good to me.

PR #812, opened 2023-09-12
Adds vendor cache gems

ajkiessl:
We need the gems stored in source to get pushed out to the servers on deploy.  I think we can stop doing this when we move to k8s since the bundling is done in the builds.


whereismyjetpack:
This was my fault. I think we are still using the cached gems in docker, but i also think that if they aren't cached they will be pulled from the internet 

I can go either way on this, the dockerfile is a bit more complex than some of our others, because we don't want the cached gems to take up space in the resultant image, so we are using an additional stage in our multi-stage build to accomplish that. 

https://github.com/psu-libraries/researcher-metadata/blob/main/Dockerfile#L23

I'd be curious to compare build times with cached vs uncached gems, I'm curious of any other benefits I might be overlooking @EricDurante ?

whereismyjetpack (approved):

ajkiessl:
I'd assume not having to do the full bundle and pull from rubygems would significantly reduce build times in most cases.  Since we're already doing the caching, let's keep it that way.

EricDurante:
@whereismyjetpack 
> I'd be curious to compare build times with cached vs uncached gems, I'm curious of any other benefits I might be overlooking?

I'd think that caching would significantly decrease build time, but apart from that I suppose that the main benefit would be eliminating dependency on the availability of rubygems.org (and any other remote sources) to build the app. They're not down often, but they are occasionally.

You could probably make an argument that caching the gems with the rest of the source codes is marginally more secure too. Unless you're always checksumming everything, every time you have to grab a copy of some code from the internet, you're exposed to a little bit of risk that somebody somewhere had a security breach and some bad actor was able to temporarily publish malicious code. If you know that you have safe code cached, then you eliminate that risk.

PR #814, opened 2023-09-12
refine pubs subject to oa workflow

### LINKED ISSUE #797: Refine the set of publications subject to the AI OA workflow
Currently, if a publication has a file from Activity Insight, RMD checks for an OA location. If the publication has an OA location, the publication does not go through the AI OA workflow.

Instead, we should check the "Open access status" field and check for a ScholarSphere open access location.

- If the "open access status" is "gold" or "hybrid", or if there is a ScholarSphere open access location, the publication should not go through the AI OA workflow. RMD should report to AI that the publication is already openly available. This information should be passed along to the AI user via the AI field that RMD calls "Activity insight postprint status." The file should either be removed from RMD or not imported.
- If the "open access status" is "green", "bronze", or "closed" and there is no ScholarSphere open access location, the publication should go through the AI OA workflow.

Smullz622:
Fixes #797


spec/component/jobs/fetch_oa_metadata_job_spec.rb
```diff
@@ -51,6 +53,12 @@
         it "updates publication's open access button last checked at" do
           expect(pub.reload.open_access_button_last_checked_at).to be_within(1.minute).of(Time.zone.now)
         end
+
+        it 'updates the publication oa status last checked at, oa status, and workflow state' do
+          expect(pub.reload.open_access_status).to eq 'closed'
```
ajkiessl, line 58: Should this be 'unknown'?

spec/component/jobs/fetch_oa_metadata_job_spec.rb
```diff
@@ -64,7 +72,8 @@
           expect(pub.reload.open_access_button_last_checked_at).to be_nil
         end
 
-        it 'updates the publication oa status last checked at and workflow state' do
+        it 'updates the publication oa status last checked at, oa status, and workflow state' do
+          expect(pub.reload.open_access_status).to eq 'closed'
```
ajkiessl, line 76: Should this be 'unknown' too?  I may be misunderstanding, but if we checked Unpaywall and didn't get anything we want to set the oa status to 'unknown', right?

app/models/publication.rb
```diff
@@ -590,6 +606,10 @@ def update_from_unpaywall(unpaywall_response)
     end
   end
 
+  def update_oa_status_from_unpaywall(unpaywall_response)
+    self.open_access_status = doi.present? || unpaywall_response.matchable_title == matchable_title ? unpaywall_response.oa_status : 'unknown'
```
ajkiessl, line 610: I think we should add a unit test for this.  The order of operations here makes me a bit uneasy, but I suspect this works.

ajkiessl (changes requested):
I have a couple of small change requests regarding the tests, but I think the functionality looks good.

spec/component/jobs/fetch_oa_metadata_job_spec.rb
Smullz622: Yep, I made a dumb mistake. Both are fixed now

ajkiessl (approved):
This LGTM.  I'm assuming the reason we don't filter OA status at the DOI step is because if we can't confirm the DOI, we can't get the OA status from Unpaywall.

Smullz622:
> This LGTM. I'm assuming the reason we don't filter OA status at the DOI step is because if we can't confirm the DOI, we can't get the OA status from Unpaywall.

Yep, that's correct

EricDurante (changes requested):
ah geez, I'm sorry - I had comments on this PR that were still pending and I didn't realize it.

I think that we need to change/add a couple of things here that weren't completely described in the issue. Please see my comments and let me know if they make sense and if you agree. Happy to discuss if you have any questions.

app/models/publication.rb
```diff
@@ -140,20 +140,30 @@ def self.preferred_versions
...
+  scope :filter_oa_status_from_workflow, -> { where.not(%{open_access_status = 'gold' OR open_access_status = 'hybrid' OR open_access_status IS NULL}) }
+  scope :needs_permissions_check, -> {
+                                    activity_insight_oa_publication
+                                      .filter_oa_status_from_workflow
+                                      .where(licence: nil, doi_verified: true, permissions_last_checked_at: nil)
+                                  }
   scope :needs_oa_metadata_search,
         -> {
           activity_insight_oa_publication
+            .filter_oa_status_from_workflow
```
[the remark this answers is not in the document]
Smullz622: Yeah, I think you're correct that filtering out NULL would create a problem, but I'm wondering if gold & hybrid should still be excluded here. When I spoke to Ana about the ticket I'm working on now, she said once something is gold/hybrid, it's permanent so I don't think we need to recheck those.

app/jobs/fetch_oa_metadata_job.rb
```diff
@@ -15,6 +15,7 @@ def perform(publication_id)
       publication.open_access_button_last_checked_at = Time.zone.now
     end
 
+    unpaywall_response.oa_status.present? ? publication.update_oa_status_from_unpaywall(unpaywall_response) : publication.open_access_status = 'unknown'
```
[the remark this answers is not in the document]
ajkiessl: Yeah, you could probably get rid of the OAB stuff here.  I guess there's the rare chance something is in OAB but not Unpaywall, in which case I think we'd want to exclude those from this workflow, though. 

I don't think we have anything built in RMD to check ScholarSphere for a single record.  The ScholarSphere import just grabs everything in ScholarSphere and sees if there's a matching DOI in RMD.  I'm pretty sure there is a DOI endpoint in ScholarSphere's API, so something could be built.

I agree we don't need all the metadata from Unpaywall for this job, so if we wanted to simplify things, we could just get whatever is necessary to validate and the open access status.  Then, leave the rest to the other importers.

app/models/publication.rb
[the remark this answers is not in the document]
EricDurante: > When I spoke to Ana about the ticket I'm working on now, she said once something is gold/hybrid, it's permanent so I don't think we need to recheck those.

@Smullz622 Ok, great - I wasn't sure about that aspect of this. So I think you're right; it definitely makes sense to still filter out gold and hybrid OA status for this scope.

app/jobs/fetch_oa_metadata_job.rb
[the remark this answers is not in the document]
EricDurante: Yeah, here's how I'd prioritize the changes that I suggested:

**High priority:**

- I think that we must check for ScholarSphere open access locations here one way or another.

**Low priority:**

- It would simplify things to remove the OAB checking/importing here since it's no longer needed, but it doesn't _hurt_ anything functionality-wise to leave it in. However, if we leave it in, then I'd say that it's high priority to ensure that we're not creating duplicate open access location records.
- Likewise, I think that the Unpaywall checking/importing is probably working correctly as is, but it looks to me like it could be made simpler either by only importing the OA status here or by just importing everything that we ultimately want from Unpaywall (OA status, OA locations, etc.) for these publications in the same way that we do elsewhere.

spec/component/jobs/fetch_oa_metadata_job_spec.rb
```diff
@@ -41,33 +44,40 @@
...
+
+      it 'does not update the publication oa data' do
+        expect(pub.reload.open_access_locations.where(source: Source::SCHOLARSPHERE)).to eq []
+      end
 
-        it 'updates the publication oa status last checked at and workflow state' do
-          expect(pub.reload.oa_status_last_checked_at).to be_within(1.minute).of(Time.zone.now)
-          expect(pub.reload.oa_workflow_state).to be_nil
-        end
+      it 'updates the publication oa status & unpaywall last checked at, oa status, and workflow state' do
```
EricDurante, line 78: The description of this block doesn't seem to quite match the actual assertions within. I think that the assertions are probably correct, but the way that the description reads is a little confusing to me. It looks like what we're actually asserting is that it "updates the OA status check timestamp and clears the OA workflow state". Should we be checking anything else?

EricDurante (changes requested):
Sorry, I didn't get a chance to scan through the tests on my last review. Your updates to the code look good to me. Just a few comments on the tests - all minor.

app/models/activity_insight_oa_file.rb
```diff
@@ -16,6 +16,7 @@ class ActivityInsightOAFile < ApplicationRecord
       .where(file_download_location: nil)
```
ajkiessl, line 16: I'm curious about this since I added it to my PR for the file version checking job.  Should we only be filtering out open access locations from ScholarSphere here now?

ajkiessl: If so, I have it in my PR and it's tested so we don't need to do it here.

EricDurante: @ajkiessl Ah, yes - that's right. Good catch.

There's no point in downloading a file for a publication that we already know to be in ScholarSphere since those publications drop completely out of the workflow as soon as we have that information, but we do want to download files for publications that have _other_ kinds of open access locations _unless_ we know that the OA status of those publications is gold or hybrid (line 19).

I didn't really think about whether or not we want to exclude files for publications that don't have an open access status yet (also line 19 here). I need to give that a little more thought when I have a minute..

spec/component/jobs/fetch_oa_metadata_job_spec.rb
Smullz622: No, I think that's all that needs to be checked & you're right about just updating the assertion.

app/models/activity_insight_oa_file.rb
Smullz622: @ajkiessl You're right. That is something that can just be resolved when it's merged right?

@EricDurante my thought process was we should exclude nulls because null means it hasn't been checked yet & could be gold/hybrid. If it has been checked & we couldn't find a status, it would be `unknown`.

ajkiessl: Yeah, my stuff should be merged soon, so as long as they're both merged before the next release we are good.

EricDurante: @Smullz622 Yeah, your reasoning totally makes sense. The thing that I was mulling over was whether or not we should go ahead and download the file anyway - even if the OA status hasn't been checked yet.

We're currently not considering whether or not we've checked for ScholarSphere OA locations (which is probably fine), we're only considering whether or not we currently know about any. We could likewise forget about considering whether or not we've checked for OA status yet and only consider whether or not it's currently gold or hybrid. Or we could go the other direction and make the file download wait until we know that we've checked ScholarSphere too.

I guess a downside of forcing the workflow to be more linear by waiting on the other checks to do the file download is that less things can potentially happen in parallel, and the whole process might take a little more time for a given publication.

A downside of allowing the file download (and dependent version checking, permissions checking, etc.) to happen in parallel with the steps for DOI checking, OA metadata checking, etc. is that we might waste a little computing power on doing those steps for publications that might get booted from the workflow later.

Another (and bigger) downside of making the file download wait for the OA metadata check is that admins won't yet have the file available to look at if the publication ends up in the manual DOI check list. DOI verification isn't currently a prerequisite for the file downloading, but it is for the OA metadata checking. So if we make OA metadata checking a prerequisite for the file download, then DOI becomes prerequisite too.  I think that having the file sometimes helps with that manual DOI verification.

Having thought this through, I'm pretty sure that we want to go ahead and download the file even if the OA status of the publication is still NULL.

Do you all agree?

EricDurante (approved):

PR #816, opened 2023-09-15
Toggle for Users to not receive OA reminders

### LINKED ISSUE #751: Create "do not email" feature to suppress OA reminder emails to specific users
In certain situations it would be nice to be able to turn off reminder emails at the level of individual users. This could be built as a toggle on/off on user pages or as a list of users that gets checked whenever emails go out. 

This feature should only suppress OA reminder emails, not any other emails sent by RMD (e.g. emails confirming waivers, emails about the proxy process, etc.).

We would not publicize this feature and would use it only in rare cases.

ajkiessl:
Adds boolean `suppress_oa_reminders` to `User`.  Allows this to be shown/edited in rails admin.  Filtering users with this set to true out of oa email notifications.

I noticed most of the test data for `#needs_open_access_notification` didn't have psu_identity data, so they weren't actually testing the criteria described in the test.  I added this to the ones that needed it.

closes #751


Smullz622 (dismissed):
lgtm!

EricDurante (approved):
LGTM

PR #817, opened 2023-09-19
Activity Insight OA Workflow version checking job

### LINKED ISSUE #717: Implement file version check job in OA Workflow
We'll need to add a scope for files that need their version checked and implement the version checking job in the OAWorkflow task.  So any file that we have a record for, and we have downloaded the file, and has no version should be a part of this scope.

This is dependent on #575.

ajkiessl:
Branched off of https://github.com/psu-libraries/researcher-metadata/pull/713 .  

I added a `version_checked` boolean to `ActivityInsightOAFile` to be a status holder like `downloaded`.  This gets set to true when the job is queued so it doesn't get queued again.

I did my best to differentiate between the ScholarSphere deposit workflow job and this.  The name of this job ended up a little weird: `AiOAWfVersionCheckJob`

closes #717


app/models/activity_insight_oa_file.rb
```diff
@@ -12,14 +12,26 @@ class ActivityInsightOAFile < ApplicationRecord
 
   mount_uploader :file_download_location, ActivityInsightFileUploader
 
-  scope :ready_for_download, -> {
+  scope :oa_type_w_no_locations, -> {
```
EricDurante, line 15: oooh, I think that this scope is going to need to be changed per #797. I guess you could either do that here or make it a separate issue and wait until after this is merged.

cc @Smullz622

ajkiessl: Ah right.  I'll do that here.

app/services/oa_workflow_service.rb
```diff
@@ -33,5 +33,14 @@ def workflow
     rescue StandardError
       file.update_column(:downloaded, false)
     end
+
+    ActivityInsightOAFile.needs_version_check.each do |file|
+      file.version_checked = true
+      file.save!
```
EricDurante, line 39: Here again, I think it might be better to do 
```ruby
file.update_column(:version_checked, true)
```
instead of
```ruby
file.version_checked = true
file.save!
```

app/services/oa_workflow_service.rb
```diff
@@ -33,5 +33,14 @@ def workflow
...
+      AiOAWfVersionCheckJob.perform_later(file.id)
+    rescue StandardError
+      file.update_column(:version_checked, false)
+      raise
+    end
```
EricDurante, line 44: I think what we're doing here is ok since we're bypassing ActiveRecord and also re-raising the original error.

app/services/oa_workflow_service.rb
```diff
@@ -33,5 +33,14 @@ def workflow
     rescue StandardError
       file.update_column(:downloaded, false)
     end
+
+    ActivityInsightOAFile.needs_version_check.each do |file|
+      file.version_checked = true
+      file.save!
+      AiOAWfVersionCheckJob.perform_later(file.id)
+    rescue StandardError
+      file.update_column(:version_checked, false)
```
EricDurante, line 42: Should we add the `version_checked` attribute to the RailsAdmin list view of the files so that we can filter for the files that are in this state?

EricDurante (changes requested):

app/jobs/ai_oa_wf_version_check_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      file.update! version: exif_file_version
+      return
+    end
+
+    pdf_file_version = FileVersionChecker.new(file_path: file.file_download_location.path,
+                                              publication: file.publication).version
+
+    file.update! version: pdf_file_version
+  rescue StandardError
+    file.update! version: 'unknown'
```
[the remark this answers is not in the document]
ajkiessl: Yeah, the error handling here was added primarily to catch malformed PDF or wrong file type errors from the PDF parser.  I'll try to make the error handling more specific.

app/services/oa_workflow_service.rb
ajkiessl: Yeah, that is a good idea.  I'll do that.

app/jobs/ai_oa_wf_version_check_job.rb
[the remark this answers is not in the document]
ajkiessl: @EricDurante I moved the error handling into the version checker class and have it only rescuing a handful of PDF::Reader errors:
https://github.com/psu-libraries/researcher-metadata/blob/7647aa7f34fdb53a48dd0dafe52ce5a45ec3fb10/app/models/file_version_checker.rb#L50-L53

ajkiessl:
@Smullz622 If you get a chance can I get a second set of eyes on the ActivityInsightOAFile scope tests and the OAWorkflowService tests?  I think I got everything in the merge.

Smullz622 (approved):
Looks good! Definitely think we need to start separating stuff out in some way though. Pretty sure I already have conflicts on my branch for 799

PR #819, opened 2023-09-20
Import unpaywall since workflow start time

whereismyjetpack:
This is a POC. One issue we have in the kubernetes environment is, some of the importers (unpaywall, and open access button) take days to complete. If the job fails due to a node eviction, maintenance, etc, the importer has to start all over the next week. Utilizing Argo Workflow retry strategies, combined with `import_since` we can more confidently get through all the publications. 

I set days_ago to 7, since we run this import weekly. this could be a rake argument that we set in the workflow. The workflow, upon fatal error, will retry with a back-off strategy. I hijacked the existing rake tasks, but I could see a scenario where we would want a days_since importer, and an all. This was just whipped up to get feedback on this idea.


ajkiessl:
Interesting concept.  I'm assuming there would be some way to pass the exact `days_ago` (or may even days and hours) based off how long the failed pod ran?

whereismyjetpack:
I think we can only pass an argument at runtime, so the thought is 'grab everything that hasn't been seen in the past 7 days and process it' If the job takes more than 7 days to run there is a fault in this logic, but given it takes less than 7 days to run I _think_ it should work out. (I will admit date math always trips me up)

That said, an operator could override that value and force it to look further back, or closer.

ajkiessl:
If we 'grab everything that hasn't been seen in the past 7 days and process it' when the job retries after failing, then we wouldn't grab anything if last week's import ran to completion.  If we only ran the import again for things older than the time the job started, then we could get all the ones that didn't run.  You could potentially also do something like: grab everything with a last checked date beyond last Monday (if the import runs on Monday).  Seems like ruby should be able to do something like that.

whereismyjetpack:
I'm able to pass a 'workflow started at' time to the job, and I'm currently testing using that datetime as a marker, so something like 

```
Publication.where(unpaywall_last_checked_at: nil..time_started_at)
```

whereismyjetpack:
I've tested this on my `djb44` branch, and It appears to be working. If the ARGO_SUBMITTED_AT time is empty, it used Time.now which, I think could catch them all.

ajkiessl (changes requested):
We'll want some tests for this.  Since we don't really need the `import_all` method anymore, we could just get rid of that, shift all of its tests over to use with `import_before`, and add a couple tests for the extra logic in `import_before`.  We'll want to test that the querying by date grabs the right things.

whereismyjetpack:
I renamed the spec to use `import_before`, and added two new specs to account for passing the date: parameter.

app/importers/unpaywall_publication_importer.rb
```diff
@@ -1,10 +1,20 @@
...
+    Rails.logger.info("Starting to import #{count} records from unpaywall")
+    Publication.where(unpaywall_last_checked_at: nil..date).or(Publication.where(unpaywall_last_checked_at: nil)).find_in_batches(batch_size: 500) do |publications|
+      publications.each do |p|
+        import_from_unpaywall(p)
+      end
+      Rails.logger.info('Processed 500 Records')
+    end
+  end
+
   def import_all
```
ajkiessl, line 15: We should delete this, and make sure it's not being used anywhere else.

whereismyjetpack: roger!

ajkiessl (approved):
lgtm

PR #820, opened 2023-09-20
Remove/change User linking code

### LINKED ISSUE #802: Remove/change temporary ActivityInsightOAFile to User linking code
Change the importer:

https://github.com/psu-libraries/researcher-metadata/blob/68e092c67b5115335d2af739fb57e16dd6be0c9e/app/importers/activity_insight_importer.rb#L205-L212

Change the model:

https://github.com/psu-libraries/researcher-metadata/blob/68e092c67b5115335d2af739fb57e16dd6be0c9e/app/models/activity_insight_oa_file.rb#L7-L11

This code should only be completed and deployed to production after the code here: https://github.com/psu-libraries/researcher-metadata/pull/801 has run for the entire Activity Insight import in production.

ajkiessl:
Makes ActivityInsightOAFile link to User required.  Removes backfill code in AI import and moves variable inside conditional.  Removes comments.

closes #802

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -202,9 +202,9 @@ class ActivityInsightImporter
           end
 
           activity_insight_file_location = pub.postprints&.first&.location
+          aif = pub_record.activity_insight_oa_files.find_by(location: activity_insight_file_location)
 
           if activity_insight_file_location.present? && pub_record.can_receive_new_ai_oa_files?
-            aif = pub_record.activity_insight_oa_files.find_by(location: activity_insight_file_location)
             if aif.blank?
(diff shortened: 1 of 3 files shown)


app/importers/activity_insight_importer.rb
```diff
@@ -202,17 +202,9 @@ def call
           end
 
           activity_insight_file_location = pub.postprints&.first&.location
-          aif = pub_record.activity_insight_oa_files.find_by(location: activity_insight_file_location)
```
Smullz622, line 205: This line will be used for the export oa status code to help get the intellcont_id & post_file_id so we probably don't want to delete it yet.

Smullz622 (approved):
lgtm!

PR #821, opened 2023-09-20
Exports oa status to Activity Insight

### LINKED ISSUE #653: Let Activity Insight know when a publication is already open access (if oa/post-print file found)
This will likely need to be done during import.  

If we detect an open access file from Activity Insight and it's associated with a publication already open access in RMD, we need to update the file dropdown in AI to indicate it's already open access.  This should be a separate job that is kicked off from the import process, though, since it will send another request that might take a few seconds to complete.

Smullz622:
Fixes #653


app/models/activity_insight_oa_file.rb
```diff
@@ -21,6 +21,14 @@ class ActivityInsightOAFile < ApplicationRecord
       .where(downloaded: nil)
       .where.not(location: nil)
   }
+  scope :send_oa_status_to_activity_insight, -> {
+                                               left_outer_joins(:publication)
+                                                 .where(publication: { publication_type: Publication.oa_publication_types })
+                                                 .left_outer_joins(publication: :open_access_locations)
+                                                 .where(open_access_locations: { publication_id: nil })
+                                                 .where(%{publication.open_access_status = 'gold' OR publication.open_access_status = 'hybrid'})
```
ajkiessl, line 29: I think we also want to send OA status if there's a ScholarSphere open access location.

app/models/activity_insight_oa_status_exporter.rb
```diff
@@ -0,0 +1,53 @@
...
+    def to_xml(file)
+      builder = Nokogiri::XML::Builder.new(encoding: 'UTF-8') do |xml|
+        xml.Data do
+          user = file.user
+          xml.Record('username' => user.webaccess_id) do
+            xml.INTELLCONT('id' => file.intellcont_id) do
+              xml.POST_FILE('id' => file.post_file_id) do
+                xml.ACCESIBLE('Already Openly Available')
+              end
+              xml.RMD_ID(file.publication.id, access: 'READ_ONLY')
```
ajkiessl, line 46: I know it says to update the `RMD_ID` in the issue, and I _think_ we can keep this, but we'll have to consider that once this value is set these records stop importing from Activity Insight.  I think this is probably okay since it's open access, but we might want to double check this with Ana or Nicole.

ajkiessl (changes requested):
I noticed a couple things which I detail below.  Also, I think we'll want to kick off these jobs in the OA Workflow rake task (`OAWorkflowService`), so we'll want to add some code for that still.

app/models/activity_insight_oa_status_exporter.rb
Smullz622: I know we discussed this, but just to document- Ana said let's update `RMD_ID` for now, and we can reassess down the road.

app/services/oa_workflow_service.rb
```diff
@@ -33,5 +33,7 @@ def workflow
     rescue StandardError
       file.update_column(:downloaded, false)
     end
+
+    AiOAStatusExportJob.perform_later
```
ajkiessl, line 37: Sorry, I should have caught this before.  I think it would be best to iterate through all the files that need their status sent to Activity Insight here, and then kick off the jobs one-by-one.  So the `ActivityInsightOAStatusExporter` would initialize with a single file rather than all the files.  Then, I think we'd want delayed_job to just handle the errors for us.  I'll describe that below.  Overall, I think doing things one-by-one will make troubleshooting and tracking jobs easier and more organized.

app/models/activity_insight_oa_status_exporter.rb
```diff
@@ -0,0 +1,55 @@
...
+        publication.save!
+      else
+        logger.error Nokogiri::XML(response.to_s).text
+        logger.error "File ID: #{file.id}"
+        not_exported_ids << file.id
+      end
+    end
+
+    logger.info "Open access status export to Activity Insight ended at #{DateTime.now}"
+    logger.info "Files not exported to AI: #{not_exported_ids}"
```
ajkiessl, line 28: Instead of doing all the logging here, we can just raise some kind of error on line 21 if we get an error response from Activity Insight.  Delayed job will then store the error and should also tell us the id of the file that had the error.

app/jobs/ai_oa_status_export_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AiOAStatusExportJob < ApplicationJob
+  queue_as 'default'
+
+  def perform
```
ajkiessl, line 6: We can pass the file id in here as an argument.  Then, `find` the file below before passing it into the exporter.

ajkiessl (changes requested):
I should have caught this earlier, but I don't think we should run all the files at once as a single job.  If they all run separately, we'll be able to keep track of errors and troubleshoot a bit easier.  Also, if delayed_job goes down for some reason, the export won't be interrupted if they are stored as separate jobs.

app/models/activity_insight_oa_status_exporter.rb
ajkiessl: Also, we'll want to set this: `publication.exported_oa_status_to_activity_insight = true` in the OAWorkflowService before queuing the job.

app/services/oa_workflow_service.rb
```diff
@@ -33,5 +33,12 @@ def workflow
     rescue StandardError
       file.update_column(:downloaded, false)
     end
+
+    ActivityInsightOAFile.send_oa_status_to_activity_insight.each do |file|
+      publication = file.publication
+      publication.exported_oa_status_to_activity_insight = true
```
ajkiessl, line 39: I think in order to cover every record in Activity Insight that needs their status updated we will have to make `exported_oa_status_to_activity_insight` an attribute of ActivityInsightOAFile.  For example: if two people upload files in Activity Insight for the same publication.  These will be two separate records in Activity Insight, but could get merged into one in RMD.  So we might end up only updating the status for one of them if we store this with Publication.

spec/component/models/activity_insight_oa_file_spec.rb
```diff
@@ -60,25 +60,34 @@
     }
     let!(:pub2) { create(:publication,
                          title: 'pub2',
-                         publication_type: 'Trade Journal Article')
+                         publication_type: 'Trade Journal Article',
+                         open_access_status: 'gold')
```
ajkiessl, line 64: I believe we are testing the exclusion of 'Trade Journal Article' from the `ready_for_download` filter with this record, so we probably want to leave the open access status as nil or 'green' (or anything that would be included).

app/models/activity_insight_oa_status_exporter.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize(file_id)
+    super()
+    @file = ActivityInsightOAFile.find(file_id)
+  end
+
+  def export
+    response = HTTParty.post webservice_url, body: to_xml(file),
+                                             headers: { 'Content-type' => 'text/xml' }, basic_auth: auth, timeout: 180
+    unless response.code == 200
```
ajkiessl, line 14: We should probably update `exported_oa_status_to_activity_insight` boolean to false here.  Also — now that I think about it — we should also add this field to the Rails Admin list and show pages.  That way admins can filter on it and see it in the show page.

ajkiessl (changes requested):
Changes looks good.  There's just a few small things I noticed.  Once those are done I think we should be good.

app/models/activity_insight_oa_status_exporter.rb
Smullz622: I added it to the rails admin pages. Question about setting `exported_oa_status_to_activity_insight` to false though- if we do that, how are we planning to handle failed exports? Anything getting exported drops out of the workflow, so if we set `exported_oa_status_to_activity_insight` as false, it won't get run through again & won't be on the radar anymore either.

ajkiessl: I think the flow would be something like:

- Export fails, flag is set to false, and an error is raised
- A developer notices the error, they or an admin can filter it in the admin console (or rails console for dev) to help troubleshoot
- Once the Error/bug is fixed, then the developer will have to manually re-run the export for that file

Outside of this export filter, these files have already fallen out of the workflow, so we shouldn't have to worry about anything else.

spec/component/models/activity_insight_oa_file_spec.rb
```diff
@@ -66,19 +66,27 @@
...
                          open_access_locations: [
                            build(:open_access_location, source: Source::OPEN_ACCESS_BUTTON, url: 'url', publication: nil)
-                         ])
+                         ],
+                         open_access_status: 'green')
     }
     let!(:pub4) { create(:publication,
-                         title: 'pub4',
-                         open_access_status: nil)
+                         title: 'pub1',
```
EricDurante, line 73: I'm guessing that you didn't mean to change these titles?

ajkiessl: Looks like pub5 and pub6 are like this too.

ajkiessl (approved):
LGTM!

PR #822, opened 2023-09-20
adds delayed_job_web

whereismyjetpack:
image

Adds delayed_job_web, mounts it at `/delayed_job` and requires `user.admin?` on that route.


ajkiessl (approved):
Looks good.  It just throws a routing error if you're not an admin.  Which seems odd to me but I guess that's fine.

EricDurante:
> It just throws a routing error if you're not an admin. Which seems odd to me but I guess that's fine.

Yeah, based on what has been added to the routes file it looks like that's because the route for the delay_job_web app/action is only defined if the current user is an admin maybe? I didn't even realize that you could change routes conditionally at runtime in this way, but I guess you can. In any case, a routing error will render as a 404 I think - which is sort of reasonable. It would  feel better to me if the request got routed either way and returned 401 if the user wasn't an admin. It looks like what we have here is how delayed_job_web is designed to work though, so it would probably be tough to change. I guess :shipit:

EricDurante (approved):

PR #826, opened 2023-09-21
File open and edit bug fix

### LINKED ISSUE #823: File opens repeatedly in AI OA Workflow
In the DOI verification workflow, if you view the file, close it, and then go into the metadata to edit it, then edit the metadata and click save, it will open another copy of the PDF and freeze the editing window so you can't make additional changes. This happens when you try to leave the DOI field blank but select "True". It's also hard to go back. Going back to the AI OA Workflow list is not enough.

ajkiessl:
Adds `target: _blank` to links to pdf downloads.  Not too sure why, but this fixes the issue.

closes #823


EricDurante (approved):
Yeah, I think it makes sense to do this regardless.

Issue #799, opened 2023-08-25
List for Publications whose file versions are all determined but Preferred Version is "None"


anaelizabethenriquez:
Related to #692 

Most of the time, if we can’t deposit the version that we got, there’s another version we can deposit (see #719, the "wrong version" bucket). But, sometimes there won’t be any version of the article that we can deposit. For these publications, an admin will set the "preferred_version" to "None". We need to add that as an option for that field. 

We don’t want those to show up in the “wrong version” bucket — they should be in a bucket of their own called "Publications for which no version can be deposited". The logic for this list is the same as the "wrong version" bucket except that "preferred_version" is "None".

In this list, an admin should be able to review the item and should be able to send the user an email explaining that no version of the publication can be deposited. This functionality can be very similar to the "wrong version" bucket, except that once we've sent an email to the user we are not waiting on any further action, so the publication(s) we emailed them about should be removed from the list. An admin should be able to send an email about just one publication or a batch of publications from a particular user.

When the publications are removed from the list (when the email is sent successfully), RMD should also:

- ~~delete the files that can't be deposited from Activity Insight (if possible)?~~
- update the publication's status in Activity Insight to "Cannot Deposit"
- delete the local copies of files that cannot be deposited from RMD and update RMD's associated database records to indicate that the files were not able to be deposited

We'll need to keep some record of the fact that we've sent this email for a given publication in order to know which publications to show in this list, but this record will also indicate that the publication has completely exited this workflow, so it will make sense to include its presence as another filter in the `Publication#activity_insight_oa_publication` scope and in any appropriate scope(s) on `ActivityInsightOAFile` so that it doesn't appear in any other lists and so that it's not subject to any more automated processing.

ajkiessl:
Update Activity Insight and send the email at the same time.

Smullz622:
@anaelizabethenriquez when you get the chance, could you add what you'd like in the email please?

anaelizabethenriquez:
Sorry about that! Here it is:

From: openaccess@psu.edu
Reply-to: openaccess@psu.edu
Subject: Open Access Post-Print Publication Files in Activity Insight

Hello,

We are writing to follow up on the publication file(s) uploaded to your account in Activity Insight in the section titled “Open Access Post-Print Publication files."

[...]

PR #829, opened 2023-09-27
Adds troubleshooting list

### LINKED ISSUE #806: Create a "troubleshooting" list in the AI OA workflow
This list should be called "Troubleshooting" and should include all publications that are subject to the workflow (#797).  We should order this by creation date, so oldest things should be at the top. 

We will use this list for troubleshooting during development and during regular use of the workflow. If publications are ending up in this list, Ana or another admin will need to review them and get them back on track.

Smullz622:
Fixes #806


app/models/publication.rb
```diff
@@ -145,6 +145,12 @@ def self.preferred_versions
                                                  .where(%{NOT EXISTS (SELECT * FROM open_access_locations WHERE open_access_locations.publication_id = publications.id AND open_access_locations.source = '#{Source::SCHOLARSPHERE}')})
                                              }
   scope :activity_insight_oa_publication, -> { oa_publication.with_no_scholarsphere_oa_locations.joins(:activity_insight_oa_files).where.not(activity_insight_oa_files: { location: nil }) }
+  scope :troubleshooting_list, -> {
+    activity_insight_oa_publication
+      .where(%{(open_access_status != 'gold' AND open_access_status != 'hybrid') OR open_access_status IS NULL})
```
EricDurante, line 150: Just a note that once the stuff that I'm working on is merged in, we'll need to add one more thing here:
```ruby
.where(%{preferred_version != '#{NO_VERSION}' OR preferred_version IS NULL})
```
In my branch, `Publication::NO_VERSION` is defined as `'None'`, which is the value that we set if we find that there are no open access permissions for either the accepted or published version of the publication. This is one more way in which the publication can completely exit the workflow. I'm not sure where/how we want to keep track of this todo item..

EricDurante: oops, scratch that ☝️. Publications with a `preferred_version` of `'None'` will technically still be in the workflow until we email the person about it (described in https://github.com/psu-libraries/researcher-metadata/issues/799). So instead of filtering on `preferred_version`, I suppose we'll need to filter on whatever new field we're going to set to indicate the the email has been sent.

Smullz622: Will whatever indicates that the email has been sent be part of the activity_insight_oa_publication scope? If so, that will cover it here already. Otherwise, if it's going to be added to steps separately, we can always just make a quick ticket.

EricDurante: > Will whatever indicates that the email has been sent be part of the activity_insight_oa_publication scope?

Yeah, I think that it would make sense for it to be part of that scope. I'll just add a note about this to #799.

app/models/publication.rb
```diff
@@ -145,6 +145,12 @@ def self.preferred_versions
...
+      .includes(:activity_insight_oa_files)
+      .order('activity_insight_oa_files.created_at ASC')
```
EricDurante, line 152: What does this do to the order of the list of publications when publications have more than one file? Does it matter?

Those aren't rhetorical questions - I don't know the answers. I just noticed that it doesn't look like any of the tests cover this scenario.

Other than that, I think this all looks good.

Smullz622: So unfortunately if there are multiple files, it's ordered by the oldest one (which is why there isn't a test). I tried to get it to push any with a newer file to the bottom but couldn't figure it out. However, when I spoke to Ana, she said she was ok with that for the time being.

PR #831, opened 2023-09-27
Give APITokens access to descendant organizations

### LINKED ISSUE #775: Remove old Organizations
We have a bunch of Organizations in RMD that are useless, they have 0 people and just make things unclear. Can we remove these? I think they came in from Pure and then Pure realized they didn't need them and they were then never removed.

ajkiessl:
Gives APITokens permission to access publications and users from the descendants of its associated organizations via the API.  The publications are filtered down to only include those that were published during the time the associated user was a part of that organization.  The users are filtered down to only include those that are _currently_ members of the organization.  Also, this gives the APIToken access to all descendant organizations from the organization API endpoint.

closes #775

### DIFF
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -5,8 +5,6 @@ class APIToken < ApplicationRecord
 
   has_many :organization_api_permissions, inverse_of: :api_token
   has_many :organizations, through: :organization_api_permissions
-  has_many :users, through: :organizations
-  has_many :publications, through: :users
 
   def all_publications
     Publication.joins(users: :organizations)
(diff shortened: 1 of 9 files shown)


app/models/api_token.rb
```diff
@@ -8,6 +8,24 @@ class APIToken < ApplicationRecord
   has_many :users, through: :organizations
   has_many :publications, through: :users
```
EricDurante, line 9:
```ruby
has_many :users, through: :organizations
has_many :publications, through: :users
```

Do we still need these two associations? Are they used elsewhere?

ajkiessl: Doesn't look like it.  I'll remove them and see what happens.

EricDurante (approved):
Looks good!

PR #832, opened 2023-09-28
789 allow either version

### LINKED ISSUE #789: Allow both Published and Accepted version as `preferred_version`

### LINKED ISSUE #790: Make License field a drop-down menu
When admins edit the "License" field, there should be a drop-down menu of options. Right now this field takes free text.

I am not sure if there is currently any validation on this field, but there should be. I believe ScholarSphere will not accept an incoming deposit if the license doesn't match an allowable license option. This is why the field is a drop-down in the regular ScholarSphere deposit workflow in RMD.

I suspect OA Button may be providing a wider range of options in this field than ScholarSphere allows, so we may need to think about mapping them (or updating ScholarSphere to allow other licenses).

EricDurante:
Whew - ok, I think this is ready to go.

This closes #789, closes #790 

Merging was pretty tough and I have a few things to note about how it went:

- There's a lot of setup (factory record creation) for the tests of the scopes on `Publication` that is now redundant and unnecessary. I left it in for now, hoping that it will make future merges easier and less confusing.
- The test coverage could also be improved for some of the new and updated scopes on `Publication`. I made note of this in some "TODO" comments.
- There's getting to be some obvious duplication in that scope logic that may be possible to refactor. I'm not sure - I haven't had a chance to try yet.

At first I liked the idea of sharing test setup for multiple ActiveRecord scopes - especially for the scopes that share a lot of logic with each other. It eliminated what would otherwise have been an enormous amount of duplicated test setup. The really big downside of doing that is that it made merging changes to that code kind of a nightmare since the scope logic and the test setup is all so complicated and intertwined. I'd be tempted to see if it would be better overall to duplicate the test setup and then deal with changing the setup in multiple places whenever shared scope logic changes - just for the sake of isolating tests and making merges easier.

I'd also be tempted to see if some other altogether different pattern works better - like "encapsulating" these scopes in their own query objects, composing those objects together to form more complicated queries, and testing them independently. I've used that pattern before in other apps that had a lot of really complex SQL query logic in the model layer, and it worked...sort of ok in terms of making testing easier. However, that's probably only because it was sort of a cheat, and overall the tests probably weren't as complete.

I've never seen a pattern for doing this kind of thing that's actually good because I don't think that there is one. The nature of integrating tightly with a relational database when you have to do a lot of complicated queries is that testing that integration thoroughly is an intractable mess.

I've worked on another app where the "database" integration was a REST API where you send queries that look like SQL over HTTP and get some JSON back representing the records that you're looking for. For that project, it wasn't really possible to have a test "database", so I ended up essentially mocking the whole "database" (REST API) for the sake of automated testing. My test setup would look like:  "allow [HTTP Client] to receive [really complicated SQL query string] and return [the JSON that we'd expect to get from that query]". That meant figuring out the correct "SQL" query via manual testing, implementing it in the API wrapper, handing it off to the "database", and trusting that as long as that SQL string (that I knew to be correct) didn't change, then the "database" would do the right thing. It didn't feel great not having any automated test at all that exercised that full integration, but if I was able to have even one simple test that went end-to-end, I'd actually feel pretty good about how the whole thing works. That situation isn't actually much different than working with the boundary between a web app and a typical relational database - it's just another external integration point. But ActiveRecord doesn't make it easy to use that same sort of pattern. With ActiveRecord, the database is inextricably part of our app, for better or worse. You can't really mock it. If you want to test it thoroughly, then you have to _integration_ test it throughly - which is a huge pain when things get complex.

Anyway, that's just another perspective to carry along as we're thinking about how to improve this stuff in the future. Maybe it's helpful - maybe not.


Smullz622 (approved):
This is intense, great work! FWIW I'm partial to the idea of isolating the test set up. It'll be a pain if a base scope changes, but trying to keep straight which pubs/files are meant for each scope is also becoming less than ideal. Separating them would make that so much easier.

ajkiessl (approved):
Looks good!  Thanks for sussing all this out.  It was a formidable redirect, but smooth landing.

Separating out the queries into their own objects might be worth a try sometime.  Then test them in isolation.  Good news is we get to do something similar again in ScholarSphere soon!  🎉

PR #833, opened 2023-09-28
Metadata review step for DM-OA deposit

### LINKED ISSUE #715: List publications that are ready for manual metadata review prior to Scholarsphere deposit
~~Once a publication has a file with a known version that we've confirmed can be deposited into Scholarsphere, we'll want to automatically check to see if the publication has all of the metadata this is required by Scholarsphere. If it doesn't, then an RMD admin will need to add or correct metadata until it's valid. However, even if the publication passes automatic validation of the metadata, an RMD admin should still review the metadata that will be sent to Scholarsphere to ensure that it actually makes sense, is formatted appropriately, etc.~~

~~So the final step in this workflow will always be manual for now, and it should consist of a list of publications that have a file that we know can be deposited. Each row in this list should show all of the metadata that's going to be sent to Scholarsphere for the given publication, and publications with missing/invalid metadata should be highlighted in some way. Each row in the list should also show a link to view/download the file that is going to be selected to be sent to Scholarsphere.~~

~~If a publication's Scholarsphere metadata is all valid, then its row in the list could include a "Deposit" button (to be implemented in a separate issue) that would send the file and metadata to Scholarsphere and update the publication's open access status/locations in RMD. If a publication's Scholarsphere metadata is not completely valid, then we could highlight the row and not display the "Deposit" button. We might also consider grouping valid and invalid publications separately in the list or even have two separate lists on the same page. In either case - valid or invalid - we should also show a link to edit the publication's metadata in the Rails Admin UI.~~

☝️The above was my initial guess at how this last step should work, but we now have a better understanding of how admins will manage this list, so we're going to do this a bit differently.

This list will still show only publications which have a file that can be deposited, but in each row of the list we'll only need to show the PSU user ID of the person who added the publication to Activity Insight, the publication's title, and the date when the most recent file for the publication was imported from Activity Insight. The list should be in order by that date so that the publications with the files that have been around the longest are at the top of the list to be worked on first.

The publication title will be a link to a detailed view of all of the publication metadata that is going to be sent to Scholarsphere along with the file. Here an admin will be able to see if any metadata is missing or incorrect prior to deposit. This page will also show a link to edit the publication metadata in the existing Rails Admin UI, and a button to deposit the file/metadata in ScholarSphere.

For now there's no need to visually indicate invalid (missing) metadata fields that are required for ScholarSphere deposit, but we should probably only display/enable the "deposit" button if all of the required metadata is present. This will hopefully prevent us from needing to handle the error condition where ScholarSphere rejects a deposit because an admin accidentally submitted incomplete metadata.

The metadata detail page should show the following information about the publication:

- title
- subtitle (if any)
- description/abstract
- DOI
- publication date
- journal title
- rights/license
- embargo date (if any)
- set statement (if any)
- a list of the names of the publication's creators/authors
- the user who will be listed as the depositor in ScholarSphere (the current RMD admin)
- a link to download the file that will be deposited

EricDurante:
Opening a PR for this feature that's still in progress..

This closes #715 when complete.


EricDurante:
I just rebased this on main and resolved conflicts.

ajkiessl:
OK this should be ready to go.  Just a note: this does not include the button, or any of the logic around the button, for sending the file off to ScholarSphere.

Smullz622 (approved):
lgtm!

PR #841, opened 2023-10-02
Fixes bug when sending single email from the wrong version list

ajkiessl:
Adds a test for this too.  The issue was with the `find` to get the publications for the email.  It returns a single record for the single email and an array for the batch email.  In the mailer, the code tries to iterate through the "publications" like an array but encounters an error when trying to iterate through the single record.  I fixed this by making the lookup for the publications a sql query that always returns an array-like object.

fixes https://github.com/psu-libraries/researcher-metadata/issues/719#issuecomment-1743690696


Smullz622 (approved):
lgtm!

PR #851, opened 2023-10-03
Adds list for when Preferred Version is None

### LINKED ISSUE #799: List for Publications whose file versions are all determined but Preferred Version is "None"

Smullz622:
Fixes #799


Smullz622:
@ajkiessl  Thinking on what Ana said at retro, should I filter pubs whose preferred version is None from this scope? It makes sense to me that we don't need to check file version if we know the preferred version is None, but I wasn't sure if I was overlooking anything. https://github.com/psu-libraries/researcher-metadata/blob/c14fde0fbf6a3223c7dc8fed001c2bc13b4219d3/app/models/activity_insight_oa_file.rb#L40

On a similar note, should pref version None be filtered from other workflow scopes? Not the base one, but scopes for each step so we aren't using resources on publications we know can't be deposited? `preferred_file_version_none_email_sent` will keep pubs out of the workflow once the email goes out, but I'm thinking about the time between when we know preferred version is none and when an admin can manually review & send the email.

ajkiessl:
@Smullz622 

> should I filter pubs whose preferred version is None from this scope? It makes sense to me that we don't need to check file version if we know the preferred version is None

Hmm my initial thought was it probably doesn't matter much to filter these out.  By the time a file record has gotten to the point where its preferred version is "None" it will have already gone through several other steps of processing, been manually reviewed, and manually selected as "None".  It will likely already have its file downloaded and have its version determined at this point.

On the other hand, if a user uploads a file, it goes through all that processing, an admin sets the preferred version to "None", _then_ the user deletes the file and uploads a new file to AI, we could avoid excess processing of that file by filtering out the "None"s.  I'm not sure it's worth making the scopes more complex to reduce a small amount of processing for rare circumstances, though.

So ultimately, I don't think it's worth filtering the "None"s out of this scope.

> should pref version None be filtered from other workflow scopes?  Not the base one, but scopes for each step so we aren't using resources on publications we know can't be deposited?

I think it depends.  I don't think admins will try to determine if the preferred version is "None" until after the preferred version has been checked in OAB.  In order to do that, we need to verify the DOI.  So we don't need to filter them out of those.  I don't think we need to filter them out of `needs_oa_metadata_search` in Publication either.  For similar reasons to the `needs_version_check` one, and I guess there's the chance we'd still want to know if something with a preferred version of "None" has any open access data.    For the `file_version_check_failed` scope in Publication, I think we do run the risk of the preferred version "None" types ending up in there when we don't want them to, so they should be filtered out of this.  The `needs_permissions_check` in ActivityInsightOAFiles is similar to the `needs_version_check` one above.  We don't _need_ to get permissions data for the files, but by the time preferred version "None" is set, it will likely have already been checked.

app/models/publication.rb
```diff
@@ -438,6 +451,7 @@ def doi_url_path
       field(:created_at) { read_only true }
       field(:updated_at) { read_only true }
       field(:updated_by_user_at) { read_only true }
+      field(:preferred_file_version_none_email_sent)
```
ajkiessl, line 454: It might make sense now to group this and "preferred_version" into a group called "Preferred Version".  It would probably also be useful to add this to the RailsAdmin "list", so admins can filter on it.

app/uploaders/activity_insight_file_uploader.rb
```diff
@@ -18,4 +19,16 @@ def cache_dir
...
+
+  def model_object_dir
+    if Rails.env.test?
+      Rails.root.join "tmp/uploads/activity_insight_file_uploads/#{model&.id || object_id}"
+    else
+      Rails.root.join "uploads/activity_insight_file_uploads/#{model&.id || object_id}"
+    end
+  end
+
+  def remove_file
```
ajkiessl, line 31: I think we should rename this to something that indicates it's removing the directory.  Technically, it's not removing the file.

spec/integration/admin/activity_insight_oa_workflow/preferred_version_none_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    it 'sends an email and displays a confirmation message' do
+      find_all("input[value='Send Email']").first.click
+      expect(page).to have_current_path activity_insight_oa_workflow_preferred_file_version_none_review_path
+      expect(page).to have_content('Email sent to abc123')
+      open_email('abc123@psu.edu')
+      expect(current_email).not_to be_nil
+      expect(current_email.body).to match(/Title 1/)
+      expect(current_email.body).not_to match(/Title 2/)
+      expect(pub1.reload.preferred_file_version_none_email_sent).not_to be_nil
+      expect(pub2.reload.preferred_file_version_none_email_sent).to be_nil
```
ajkiessl, line 79: I think we should test what happens with the `AiOAStatusExportJob` here; like the test above.

ajkiessl: Also, I think we should test the contents of the email in `faculty_notifications_mailer_spec.rb` as a unit test.  We still want to test that an email was sent here in `preferred_version_none_spec`, but not the contents.

ajkiessl (changes requested):
I have a couple thoughts on some things below 👇

spec/integration/admin/activity_insight_oa_workflow/preferred_version_none_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(pub2.reload.preferred_file_version_none_email_sent).not_to be_nil
+      expect(AiOAStatusExportJob).to have_received(:perform_later).with(aif1.id, 'Cannot Deposit')
+    end
+  end
+
+  describe 'clicking the button to send a single email notification' do
+    before { allow(AiOAStatusExportJob).to receive(:perform_later) }
+
+    it 'sends an email and displays a confirmation message' do
+      find_all("input[value='Send Email']").first.click
```
ajkiessl, line 72: We'll want to keep an eye on this one.  It could start failing like the other one.

ajkiessl (approved):
Looks good!

PR #853, opened 2023-10-04
Removes oa metadata fields from publication

### LINKED ISSUE #830: Remove OA permissions metadata fields from the Publication model
The `Publication` model/database table currently has fields for:

- `licence`
- `set_statement`
- `embargo_date`
- `checked_for_set_statement`
- `checked_for_embargo_date`

These fields have been added to the `ActivityInsightOAFile` model/table and should be removed from the `Publication` model since they are no longer used there.

Smullz622:
Fixes #830

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -418,7 +418,6 @@ class Publication < ApplicationRecord
         end
         field(:doi_verified)
       end
-      field(:preferred_version)
       field(:activity_insight_postprint_status)
       field(:open_access_status)
       field(:open_access_button_last_checked_at)
@@ -447,6 +446,7 @@ class Publication < ApplicationRecord
       field(:imports)
       field(:organizations)
       field(:visible) { label 'Visible via API?' }
+      field(:preferred_version)
     end
 
     edit do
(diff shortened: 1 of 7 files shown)


app/models/publication.rb
```diff
@@ -421,14 +418,7 @@ def doi_url_path
...
       end
-      group :open_access_permissions do
-        field(:preferred_version)
-        field(:set_statement)
-        field(:checked_for_set_statement)
-        field(:licence)
-        field(:embargo_date)
-        field(:checked_for_embargo_date)
-      end
+      field(:preferred_version)
```
ajkiessl, line 421: I think this is fine, but it might be a bit more user friendly if the :preferred_version was down at the bottom near the DOI grouping.  Just so all the AI OA Workflow stuff is close together.

db/schema.rb
```diff
@@ -495,15 +495,11 @@
...
-    t.string "licence"
-    t.date "embargo_date"
-    t.string "set_statement"
     t.string "preferred_version"
     t.datetime "permissions_last_checked_at"
     t.datetime "oa_status_last_checked_at"
-    t.boolean "checked_for_set_statement"
-    t.boolean "checked_for_embargo_date"
     t.datetime "wrong_oa_version_notification_sent_at"
+    t.boolean "preferred_file_version_none_email_sent"
```
ajkiessl, line 502: This probably won't cause any issues, but technically it shouldn't be here.

ajkiessl (changes requested):
Just a couple minor things 👇 .

ajkiessl (approved):
LGTM!

PR #856, opened 2023-10-04
Adds notArticleFile to file version options

### LINKED ISSUE #840: Version option for files determined to be neither the Accepted Version nor the Published Version
Currently, the "Version" field has three possible values: "unknown", "acceptedVersion", and "publishedVersion". Sometimes, Activity Insight users upload files that are not article files. We should add another possible value to the "Version" field: "notArticleFile". 

Eventually, such files should connect to #839. For MVP, we just need a way for reviewers to assign these a version so they don't clog up the "Needs Version Review" list.

Smullz622:
Fixes #840

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -232,3 +232,5 @@ en:
     published_version_display: 'Final Published Version'
     published_or_accepted_version_display: 'Final Published Version or Accepted Manuscript'
     no_version_display: 'None'
+    not_article_file: 'notArticleFile'
+    not_article_file_display: 'Not Article File'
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -232,3 +232,5 @@ en:
     published_version_display: 'Final Published Version'
     published_or_accepted_version_display: 'Final Published Version or Accepted Manuscript'
     no_version_display: 'None'
+    not_article_file: 'notArticleFile'
```
ajkiessl, line 235: Like Eric had mentioned in one of my PRs: we should stop doing this.  The `not_article_file_display` is fine, but `not_article_file` should be a constant defined in ActivityInsightOAFile.  Something like `NOT_ARTICLE_FILE = 'notArticleFile'`.  Then in `ALLOWED_VERSIONS` use `NOT_ARTICLE_FILE` instead of 'notArticleFile' and in the version_status_display method use `NOT_ARTICLE_FILE` instead of the I18n call.

ajkiessl (changes requested):
Looks good except for one thing.

config/locales/en.yml
Smullz622: Right, I forgot about that. Thanks!

ajkiessl (approved):

PR #862, opened 2023-10-05
Add DOI Error field to Publication

### LINKED ISSUE #844: Add a DOI_Error Boolean and exclude Publications with DOI_Error = True from the DOI verification "bucket"
- Add another Boolean in the DOI section of the Publication metadata, called something like "DOI_Error".
- Update the scope for the DOI verification bucket so that Publications with DOI_Error = True are excluded from the manual verification list. (Such Publications should still be included in automated DOI checking, to the extent that that process repeats periodically.)
- Enable admins to filter on this field in the admin console, so we can find these Publications for periodic re-review.

Use case: Admins conducting manual DOI verification will set this field to "True" if the publication is not yet published (and therefore we don't know yet if it will have a DOI) or if the DOI does not resolve (in which case we report the error to Crossref and need to wait for it to get fixed). In both these situations, we need to know that we've already reviewed the record and remove it from the regular DOI verification list. But, we also need to come back eventually and review it again.

Smullz622:
Fixes #844


spec/integration/admin/activity_insight_oa_workflow/wrong_file_version_review_spec.rb
```diff
@@ -69,7 +69,9 @@
 
   describe 'clicking the button to send a single email notification' do
     it 'sends an email and displays a confirmation message' do
+      sleep(2)
```
ajkiessl, line 72: If the sleep here did fix the test, then I think `sleep(0.5)` will probably be sufficient.  Also, I think the second sleep below can be removed.  At that point capybara is just using whatever is stored in `inputs` to determine what to click, so there shouldn't be anymore delays causing trouble.

ajkiessl: I don't think the sleep is going to be reliable.  I managed to wrack my brain enough to put together an algorithm that should get it right every time:

`find_all("input[name='publications']", visible: false).collect{|i| i if i.value == pub1.id.to_s}.compact.first.sibling("input").click`

ajkiessl: Or as rubocop prefers:

`find_all("input[name='publications']", visible: false).select { |i| i.value == pub1.id.to_s }.first.sibling('input').click`

ajkiessl (approved):
:shipit:

ajkiessl (approved):

PR #867, opened 2023-10-09
Hotfix:  `distinct` AI File scopes

ajkiessl:
It appears the double `left_outer_joins` cause duplicates in the `ActivityInsightOAFile` scopes.  I added `distinct`s to the end of the scopes to get rid of the duplicates.  I also added some oa locations to some of the test fixtures to test that the duplicates are removed by the `distinct`.

I'm not sure if there's another way to fix this.  Changing one, or both, of the outer joins to regular joins doesn't fix it.


Smullz622 (approved):
LGTM!

PR #868, opened 2023-10-09
deletes files no longer in activity insight

### LINKED ISSUE #858: During import, delete files in RMD that no longer have a file in Activity Insight
Now that we are collecting the `intellcont_id`s from Activity Insight and storing them with ActivityInsightOAFiles, we'll be able to tell, on import, when a file is no longer in Activity Insight.  The `intellcont_id` should match the `source_identifier` with the PublicationImport, so when we get to an import record and see that it has no file to be imported _but_ we have a file stored with that `intellcont_id`, that file can be deleted.

Smullz622:
Fixes #858

### DIFF
--- a/app/models/activity_insight_oa_file.rb
+++ b/app/models/activity_insight_oa_file.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ActivityInsightOAFile < ApplicationRecord
+  after_destroy :remove_file_download_location!
+
   NOT_ARTICLE_FILE = 'notArticleFile'
   ALLOWED_VERSIONS = [
     I18n.t('file_versions.accepted_version'),
--- a/spec/component/importers/activity_insight_importer_spec.rb
+++ b/spec/component/importers/activity_insight_importer_spec.rb
@@ -4993,6 +4993,27 @@ describe ActivityInsightImporter do
                                       publication: existing_pub) }
       let!(:existing_pub) { create(:publication) }
 
+      context 'when the publication to be imported does not have a file' do
+        let!(:existing_import2) { create(:publication_import,
+                                         source: 'Activity Insight',
+                                         source_identifier: '190706413568',
+                                         publication: pub1) }
+        let!(:aif1) { create(:activity_insight_oa_file,
+                             publication: pub1, version: 'publishedVersion',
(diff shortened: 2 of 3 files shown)


app/models/activity_insight_oa_file.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ActivityInsightOAFile < ApplicationRecord
+  after_destroy :remove_file_download_location!
```
ajkiessl, line 4: I think it would be a good idea to test the removal of the file and the file_download_location metadata  within the `ActivityInsightOAFile` tests now that this is here.

spec/component/importers/activity_insight_importer_spec.rb
```diff
@@ -4993,6 +4993,27 @@
...
+                                         publication: pub1) }
+        let!(:aif1) { create(:activity_insight_oa_file,
+                             publication: pub1, version: 'publishedVersion',
+                             file_download_location: fixture_file_open('test_file.pdf'),
+                             intellcont_id: '190706413568') }
+        let!(:pub1) { create(:publication) }
+
+        let!(:file_download_directory1) { aif1.file_download_location.model_object_dir }
+
+        it 'deletes the ActivityInsightOAFile and downloaded file' do
```
ajkiessl, line 5009: I think we should test and describe the two potential outcomes here in their own contexts.  At least so the behavior is documented.  

- if an ActivityInsightOAFile exists: it gets deleted
- if an ActivityInsightOAFile does not exist: nothing happens

ajkiessl (changes requested):
I have a couple of suggestions for this.

app/models/activity_insight_oa_file.rb
Smullz622: I added a test for the removal of the file directory, but I'm not sure how to test for the metadata since it's attached to the `ActivityInsightOAFile` & that's being destroyed. Please let me know if I'm missing something

ajkiessl: Ah whoops, that's right.  I don't think there would be a way to test to removal of the metadata.

ajkiessl (dismissed):

ajkiessl (approved):

PR #869, opened 2023-10-10
manual version review display bug fix

### LINKED ISSUE #846: Display for Manual Version Review gives links in two places
Publications that have more than one file associated with them are showing up oddly in the [Manual Version Review list](https://metadata-qa.libraries.psu.edu/activity_insight_oa_workflow/file_version_review). Each file is showing up in its own line, which is what we want, with a link to edit the metadata for the file. But links to both files show up in that row. That means for a publication with two files, four links to the files appear.

I'm attaching a screenshot illustrating this. I will try to leave this live on QA for this publication.
Screenshot 2023-10-03 at 12 26 29 PM

If this is very difficult to fix, we might want to hold off given that it impacts a small number of publications. But with this bug it would be very easy for an admin to edit the Version data for the wrong file.

Smullz622:
Fixes #846

### DIFF
--- a/app/views/activity_insight_oa_workflow/file_version_curation/index.html.erb
+++ b/app/views/activity_insight_oa_workflow/file_version_curation/index.html.erb
@@ -23,15 +23,14 @@
                       rails_admin.edit_path(model_name: :activity_insight_oa_file, id: file.id),
                       target: '_blank' %>
                 </td>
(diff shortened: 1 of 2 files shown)


ajkiessl (approved):
Looks good.  Thanks!

PR #872, opened 2023-10-10
sets postprint_status to 'In Progress' on import

### LINKED ISSUE #871: Set post_print_status to "In Progress" on import
We want the Activity Insight Importer to update the post_print_status in RMD to "In Progress" and kick off a job that updates the status to "In Progress" in Activity Insight when it gets a new file.

We'll have to come up with a way to back-fill this.  We should be careful here, since some of the things we've already imported could have a status of "Already Openly Available".  We really only want to update the status to "In Progress" for anything that has a blank status in Activity Insight.

Something I think we need to reconsider is how we import records from Activity Insight that have an `RMD_ID`.  We could just work around the logic in the status exporter to not update the `RMD_ID`, but I'm beginning to think we don't need to skip importing things with `RMD_ID`s anymore.  That logic was added when we didn't have very good duplicate auto-mergers.  The way things work now, if we were to import a publication from Activity Insight that was previously exported from RMD, then it would just create an nearly identical duplicate that would subsequently be auto-merged when the auto-mergers run.  The end result would just be that we end up with an import record pointing us to the record in Activity Insight.  I consider this a net positive.  So, I think we also remove the logic in the importer that skips over Activity Insight records with `RMD_ID`s in this issue.  Then we don't have to worry about that interfering with any of our future Activity Insight exports.

Smullz622:
Fixes #871


spec/component/importers/activity_insight_importer_spec.rb
```diff
@@ -5029,6 +5031,7 @@
 
           it 'creates a new ActivityInsightOAFile for that publication' do
             expect(DOIVerificationJob).to receive(:perform_later).exactly(3).times
+            expect(AiOAStatusExportJob).to receive(:perform_later).once
```
ajkiessl, line 5034: I think we should have a test somewhere in here that tests when the status doesn't update because it is not blank.  We should also test that the post_print_status gets updated in the database.  This would be just for when there's a new file.  I'm not too concerned about tests for the back fill code since it will be removed.

ajkiessl (changes requested):
This looks good.  I have just one change request.

PR #873, opened 2023-10-10
AI OA Workflow ScholarSphere deposit

### LINKED ISSUE #716: Allow publications to be deposited into Scholarsphere from the last step/list in the new workflow
This implements the "Deposit" button described in https://github.com/psu-libraries/researcher-metadata/issues/715. Clicking this button will result in essentially the same behavior as submitting the Scholarsphere self-deposit form that already exists in RMD, so we should be able to reuse a lot of the same code/functionality.

If the publication has more than one associated file upload and only one is the version that can be deposited, then it should select that one file, or, if there is more than one file that is the version that can be deposited, then it should select one of those files arbitrarily. The selected file, along with the required metadata from the publication record should then be uploaded to Scholarsphere in the background. After the upload has succeeded, the publication's open access status and location information should be updated in RMD which will cause the publication to no longer appear in any of the lists and to no longer be picked up by any of the background jobs in this workflow. At least for now, the "Deposit" button should trigger a confirmation dialog.

ajkiessl:
Implements the button to send the publication metadata and its file over to ScholarSphere.

closes #716

### DIFF
--- a/app/controllers/activity_insight_oa_workflow/metadata_curation_controller.rb
+++ b/app/controllers/activity_insight_oa_workflow/metadata_curation_controller.rb
@@ -7,5 +7,36 @@ class ActivityInsightOAWorkflow::MetadataCurationController < ActivityInsightOAW
 
   def show
     @publication = Publication.ready_for_metadata_review.find(params[:publication_id])
+  rescue ActiveRecord::RecordNotFound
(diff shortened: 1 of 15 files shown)


Smullz622 (approved):
LGTM!

PR #883, opened 2023-10-16
add screen reader improvements to user's pub list

### LINKED ISSUE #610: Add alt text for icons in the user's Publications list
The icons in the key and the publications list are getting missed by screen readers. Add alt text for what they mean, in the table of publications. Hide the icon key from a screen reader user.

Smullz622:
Fixes #610

### DIFF
--- a/app/decorators/authorship_decorator.rb
+++ b/app/decorators/authorship_decorator.rb
@@ -41,7 +41,7 @@ class AuthorshipDecorator < BaseDecorator
     when 'exclamation-circle'
       'Scholarsphere upload failed. Please try again'
     when 'circle-o-notch'
-      'not be subject to the open access policy until published'
+      'publication is in press and will not be subject to the open access policy until published'
     else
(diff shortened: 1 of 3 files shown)


app/decorators/authorship_decorator.rb
```diff
@@ -30,6 +30,23 @@ def open_access_status_icon
...
+    when 'unlock-alt'
+      'known open access version'
+    when 'lock'
+      'Open access obligations waived'
+    when 'hourglass-half'
+      'Upload to ScholarSphere pending'
+    when 'exclamation-circle'
+      'Scholarsphere upload failed. Please try again'
+    when 'circle-o-notch'
+      'not be subject to the open access policy until published'
```
ajkiessl, line 44: I think it would be good to mention that the publication is "In Press" here.  Maybe just add "publication is in press and will..." to the beginning of this?

ajkiessl (changes requested):
Just one small request here:

ajkiessl (approved):
Looks good, thanks!

PR #887, opened 2023-10-17
adds header to file version check & author to troubleshooting table

### LINKED ISSUE #608: Add heading on version check page
The version check page (which gives the user a choice between accepted version and published version) has no heading beyond Heading Level 1. There should be a heading (presumably a Heading 2?) right before the “we have not been able to determine...” text.

### LINKED ISSUE #886: Add column with user AccessID in Troubleshooting bucket
In the Troubleshooting bucket of the AI OA Workflow, add a column listing the Access ID of the user associated with the file. This will enable admins to find-on-page to see quickly if we have files from a given user.

Smullz622:
Fixes #608, Fixes #886

I accidentally pushed changes for #886 to this branch, but they're both small enough changes, I figured it be easier to just leave them together.

### DIFF
--- a/app/views/activity_insight_oa_workflow/all_workflow_publications/index.html.erb
+++ b/app/views/activity_insight_oa_workflow/all_workflow_publications/index.html.erb
@@ -4,6 +4,7 @@
   <table class="table">
     <thead class='card-header'>
       <tr>
+        <th>Author</th>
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
Looks good to me.  Thanks!

PR #889, opened 2023-10-17
README Updates 10/17

ajkiessl:
Removes developer notes.  I moved these to the wiki.  Adds some information about the Activity Insight Open Access workflow.


Smullz622 (approved):
lgtm!

PR #895, opened 2023-10-19
Adds offset and limit query params to organization publications endpoint

### LINKED ISSUE #893: Pagination for organization's publications API endpoint
The College of Medicine recently tried pulling all publications for their organization via the API.  They received a 504 time-out error response instead of the json of publications.  COM has 52,000+ publications.  The system cannot handle this large of a response.

### Solution

Implement pagination for this endpoint to allow smaller responses.  There should be optional query parameters for page size and number.  API users should still be able to grab all publications at once if they'd like.

ajkiessl:
closes #893 

I mention in the issue to do pagination.  I used the offset and limit parameters instead to give end users a little more flexibility.  I also think this is more commonly used with APIs.

### DIFF
--- a/app/controllers/api/v1/organizations_controller.rb
+++ b/app/controllers/api/v1/organizations_controller.rb
@@ -8,7 +8,9 @@ module API::V1
 
     def publications
       org = api_token.all_organizations.visible.find(params[:id])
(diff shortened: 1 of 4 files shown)


Smullz622 (approved):
lgtm

PR #900, opened 2023-10-23
Removes backfill

### LINKED ISSUE #898: Remove backfill of "In Progress" status
https://github.com/psu-libraries/researcher-metadata/pull/872/files#diff-eaf30dbedae55e39fb1b8657b9976373e140cd74e128428853b82a8f29f626e8R206-R213

ajkiessl:
Do not merge until the Activity Insight import has run in production after the recent release.

closes #898

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -205,7 +205,10 @@ class ActivityInsightImporter
 
           if activity_insight_file_location.blank?
             existing_file = ActivityInsightOAFile.find_by(intellcont_id: pub.activity_insight_id)
-            ActivityInsightOAFile.destroy_by(intellcont_id: pub.activity_insight_id) if existing_file.present?
+            if existing_file.present?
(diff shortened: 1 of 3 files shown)


Smullz622 (approved):
lgtm!

ajkiessl:
@Smullz622 @anaelizabethenriquez I spent a good part of the morning trying to figure out why there were still so many publications in RMD with an Activity Insight file, but no postprint status after the AI import last night.  I finally realized it's because of duplicates in Activity Insight.  Say we import a publication's metadata and it has a file and postprint status.  Then, later in the import, we import a duplicate of that publication and it has no file or postprint status.  This results in the postprint status being set back to nil (empty value).  I updated the AI importer to only import the postprint status if the postprint status is present.

The only way to set the postprint status back to nil in Activity Insight is by deleting the file.  So, I also added to the AI importer some code to set the postprint status in RMD to nil when we delete a file.

anaelizabethenriquez:
@ajkiessl Sounds good; thanks for the heads up!

Smullz622 (approved):
lgtm!

PR #901, opened 2023-10-26
Rails admin upgrade from v2 to v3

### LINKED ISSUE #494: Activity Insight publication exporter (by organization) not working
If you try exporting a publication/s to Activity Insight's beta environment, you'll see nothing exports.  It doesn't even leave an error message in the logs. 

I would start by trying to rewrite the AiPublicationExportJob to imitate the ScholarsphereUploadJob and see if that fixes it.  The ScholarsphereUploadJob inherits from ApplicationJob and it has a defined queue (AiPublicationExportJob does not).  These things might be causing the issue.

It is being queued by delayed job, but no error is shown.  The 'attempts' column is 0, so I'm assuming it is not running.

### LINKED ISSUE #899: rmd security vulnerabilities

ajkiessl:
This clears up the jquery-ui vulnerability since RailsAdmin 3 does not use jquery-ui.  The Rails vulnerability is also fixed in here.

A couple things to note:
- I also removed a lot of unused capistrano stuff we don't need now that we use Argo and stuff to deploy
- We were using a gem called rails_admin_toggleable to allow admins to toggle some booleans from the list views.  I don't think this was really being used.  Regardless, it had to be removed because it is not maintained and not compatible with RailsAdmin 3.
- RailsAdmin 3 has CSS that needs bundling, so `css-bundling` was added.  It works similarly to `js-bundling`.
- We have a lot of custom actions and views that are basically copy-pasted from the RailsAdmin source code and slightly modified for our needs.  I had to go through and update the code in those to match the RailsAdmin 3 syntax.
- Another ancillary thing in this PR is: I fixed the Activity Insight Publication Exporter.  I figured I'd fix that since I had to update some of the code for that action.
- Zeitwerk/autoloading did not like this upgrade.  The only way I could get this to work was by wrapping the RailsAdmin configuration initialization in a `to_prepare` block.  I have a comment about that in the initializer.

Preview: https://researcher-metadata-rails-admin.dev.k8s.libraries.psu.edu/
closes #899 
closes #494


Smullz622 (approved):
LGTM, great work!

PR #923, opened 2023-12-18
Replace return with next and message to stdout

ajkiessl:
`return` causes an error in rake tasks.  Using `next` and a message to STDOUT instead.


Smullz622 (approved):
lgtm

PR #925, opened 2024-01-02
Turn off Turbolinks on duplicate merge form

ajkiessl:
The popup when manually de-duping publications and the publications are in a non-duplicate group was not popping up.  This was because Turbolinks in the new bundled RailsAdmin assets was causing the page to render rather than reload.  I fixed this by turning off Turbolinks in that form.  I turned on javascript for the tests for this feature to test that Turbolinks doesn't interfere.


Smullz622 (approved):
lgtm

PR #929, opened 2024-01-11
Vulnerability Updates

### LINKED ISSUE #928: rmd gems

ajkiessl:
closes #928 

This one was pretty straight forward.  

For the view_components upgrade, a method became obsolete in the test suite that I had to change. 

Carrierwave: I did not upgrade to the next major version.  There was a patch for the current version.  The major version broke things.

The 'webpack-jquery-ui' node module is not being maintained.  I removed that and am just using 'jquery-ui' (which is already a dependency for RailsAdmin).


Smullz622 (approved):
lgtm!

PR #934, opened 2024-01-24
DOI not required for submission to ScholarSphere

### LINKED ISSUE #933: Cannot deposit publication without a DOI from the last step of AI OA Workflow
It doesn't seem to be possible to deposit a file from the last step in the AI OA Workflow if the publication does not have a DOI. If there is no DOI, "Insufficient metadata to upload to ScholarSphere" is displayed. I don't believe this is how we intended for this to work, but I'm also not sure that we tested this with a publication that doesn't have a DOI. 

Here is an example: https://metadata.libraries.psu.edu/activity_insight_oa_workflow/metadata_review/460659

For publications that have DOIs, we want to make sure that we find and include the DOI with the deposit. But some publications do not have DOIs at all, so if the "DOI verified?" Boolean is set to "True", we should be able to deposit even if the DOI field is blank.

ajkiessl:
Removed requirement for DOI to submit a file to ScholarSphere from the AI OA Workflow

closes #933

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -744,7 +744,6 @@ class Publication < ApplicationRecord
       title.present? &&
       abstract.present? &&
       published_on.present? &&
-      doi.present? &&
       doi == DOISanitizer.new(doi).url &&
       !scholarsphere_upload_pending? &&
       !scholarsphere_upload_failed?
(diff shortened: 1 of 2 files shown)


Smullz622 (approved):
lgtm!

Issue #755, opened 2023-06-13
Duplicate deposits to ScholarSphere


anaelizabethenriquez:
Paulina has reported that ScholarSphere is getting duplicate deposits of articles from RMD. The example I have right now is RMD publication 533969, which has the following two deposits:

- https://scholarsphere.psu.edu/resources/67450b81-6bb1-4d00-befb-6dd73a3015c2
- https://scholarsphere.psu.edu/resources/cd1410e4-9bac-44c5-8a3b-5e163d06b451

That publication has only one import associated with it, so I don't think it's the product of a duplicate merge. It's not in a duplicate group now. And I don't see any other records in RMD that look like they ought to be merged with it.

Both ScholarSphere records are linked from this publication. I don't understand how the user was able to deposit the article a second time, since it was already in ScholarSphere.

I don't have any other examples right now, but Paulina said there are several of these. I'll update this issue if I learn about more.

ajkiessl:
Okay, I figured out the rest of it.  Judging by the ScholarSphere ingest endpoint code, we aren't stopping the creation of a record if there's a PsuIdentity::SearchService::NotFound error like we encountered with the first example above.  I can add an issue to the ScholarSphere backlog to handle this differently.  We should either be returning a 200, creating a record, and then returning some kind of warning message, _or_ returning a 400 and not create the record.

For last example above, those are in fact work:work_version pairs.  These must've gotten around our attempts to clean these out of our OpenAccessLocations.  Not too sure how, though.

anaelizabethenriquez:
Thanks @ajkiessl! Sounds like with the issue you added to the ScholarSphere backlog, we should be covered on this. Looping in @PaulinaKrys so she can see all this. Paulina, if you see any more examples of this, please let me know.

PaulinaKrys:
Thanks @ajkiessl and @anaelizabethenriquez! I actually went through some more deposits and found a couple of additional example pairs for you to take a look at. All cases appear to be more similar to the last example from earlier where the metadata varied between submissions (formatting of the author names, etc.):

- https://scholarsphere.psu.edu/resources/67914f4d-88cd-4a54-9f8a-6856075507a2 
- https://scholarsphere.psu.edu/resources/072d850d-8e56-414f-a090-2df98c7a48f1

- https://scholarsphere.psu.edu/resources/c1ff2067-750c-438d-95e8-b7648d0c07f1
[...]

PR #938, opened 2024-01-31
Fix double depositing scholarsphere deposits in AI workflow

ajkiessl:
Adds controller level check that the file can be deposited.  This will prevent double clicks from creating multiple deposits.

fixes this comment https://github.com/psu-libraries/researcher-metadata/issues/755#issuecomment-1917971680

### DIFF
--- a/app/controllers/activity_insight_oa_workflow/metadata_curation_controller.rb
+++ b/app/controllers/activity_insight_oa_workflow/metadata_curation_controller.rb
@@ -14,6 +14,10 @@ class ActivityInsightOAWorkflow::MetadataCurationController < ActivityInsightOAW
 
   def create_scholarsphere_deposit
     publication = Publication.find(params[:publication_id])
+    unless publication.can_deposit_to_scholarsphere?
+      flash[:warning] = I18n.t('activity_insight_oa_workflow.metadata_curation_list.cannot_be_deposited')
+      return redirect_to activity_insight_oa_workflow_metadata_review_path
+    end
     activity_insight_oa_file = publication.ai_file_for_deposit
     authorship = Authorship.find_by(user: activity_insight_oa_file.user,
                                     publication: publication)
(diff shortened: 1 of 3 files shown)


app/controllers/activity_insight_oa_workflow/metadata_curation_controller.rb
```diff
@@ -14,6 +14,10 @@ def show
 
   def create_scholarsphere_deposit
     publication = Publication.find(params[:publication_id])
+    unless publication.can_deposit_to_scholarsphere?
+      flash[:warning] = I18n.t('activity_insight_oa_workflow.metadata_curation_list.cannot_be_deposited')
```
Smullz622, line 18: When you were testing this, how did it go timing wise with the success notice? If someone double clicks, will the success pop up over the warning or will it be left on the warning?

ajkiessl: The warning shows up.

Smullz622: Assuming they double click fairly quickly, the first click creates a successful deposit, and the second click triggers the warning, is it going to make users think that their deposit was entirely unsuccessful? Or am I missing something?

ajkiessl: Yeah, I'm conflicted on the wording too.  Technically, this could happen in more instances than just a double-click.  For instance if someone opened another tab for the submission form, then edited some of the metadata for the publication (rendering it no longer depositable), then we'd want this validation to stop that too.  This flash message is just very general for a user mistake that's hard to code for.

ajkiessl: So yes, the user will initially think it didn't submit, but if they check the scholarsphere deposits tab in the admin dashboard they'll see it deposited.  Also, they'll see it in scholarsphere.

Smullz622: Yeah, that is tricky. If the double click isn't happening that often, we're probably good leaving it general, but if it's more frequent, we might want to consider adding something more in the warning text or even something preemptive by the button saying click once then wait

ajkiessl: Agreed

Smullz622 (approved):

PR #941, opened 2024-02-02
Prevent double-clicks

ajkiessl:
Turns out what I implemented earlier doesn't prevent double submissions on double-click in a live environment.  Since the deposit happens quickly on QA, the work deposit's status becomes 'Success' while the user can still click.  If the user clicks after the status becomes 'Success', it will submit again.

I implemented two things to help with this.  One is to only pull from publications that are `ready_for_metadata_review` in the controller.  If a deposit is successful, it should fall out of this scope since it will have a scholarsphere url.  As another layer of protection, I modified the deposit button to disable after being clicked.


ajkiessl:
actually fixes https://github.com/psu-libraries/researcher-metadata/issues/755#issuecomment-1917971680

PR #990, opened 2024-07-03
Fix broken ORCID test

ajkiessl:
Instead of testing that the response message is 'Created' just check for the 'Created' response code (201).  The api must've changed to stop sending the message.

This was just an issue with the test.  The actual ORCID client uses the response code in its logic.


Smullz622 (approved):
LGTM

PR #1019, opened 2024-11-08
Updates all around

jrpatterson:
CI:  Updated docker engine version and ruby orbs
Revised dockerfile based on Scholarsphere
Added docker-compose env to run app locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM harbor.k8s.libraries.psu.edu/library/ruby-3.1.2-node-16:20240701 AS base
 # Isilon has issues with uid 2000 for some reason
-# change the app to run as 201
+# change the app to run as 201, or as your local user id (1000 - jrp22)
 ARG UID=3000
 
 USER root
@@ -38,7 +38,7 @@ USER root
 RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list
 
-    # hadolint ignore=DL3008
+# hadolint ignore=DL3008
 RUN apt-get update && apt-get install -y --no-install-recommends\
   rsync \
   google-chrome-stable \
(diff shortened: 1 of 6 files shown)


lauterman (dismissed):
docker compose up failed for me:

Exiting:  /usr/local/lib/ruby/3.1.0/fileutils.rb:243:in `mkdir': Permission denied @ dir_s_mkdir - /app/tmp/cache (Errno::EACCES)

config/database.yml
```diff
@@ -10,7 +10,7 @@ common: &common
 
 development:
   <<: *common
-  database: <%= ENV.fetch("POSTGRES_DB", Settings.postgres.database ) %>_development
+  database: <%= ENV.fetch("POSTGRES_DB", Settings.postgres.database ) %>
```
ajkiessl, line 13: Can this be changed back to have the "_development" on the end?  That's the standard for rails.  This change will probably break our development environments.

Dockerfile
```diff
@@ -1,25 +1,20 @@
-FROM harbor.k8s.libraries.psu.edu/library/ruby-3.1.2-node-16:20240701 as base
+FROM harbor.k8s.libraries.psu.edu/library/ruby-3.1.2-node-16:20240701 AS base
 # Isilon has issues with uid 2000 for some reason
-# change the app to run as 201
-ARG UID=201
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+# change the app to run as 201, or as your local user id (1000 - jrp22)
```
ajkiessl, line 3: Should this say 3000?

ajkiessl (changes requested):
Looks good.  I noticed two things.

ajkiessl:
Also, fwiw I was able to build and use docker compose up to run the web and db containers.  I ran the db migrations from within the container, and the app is mostly working.  We'll need to get the environment variables loaded into the web container to see if the rest of the app works.  That can be done at a later time, though.

Dockerfile
jrpatterson: I was having some local issues and those are resolved.  I set this back to what it was prior (201)

Dockerfile
```diff
@@ -1,25 +1,20 @@
...
 # change the app to run as 201
 ARG UID=201
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
+USER root
 WORKDIR /app
-
-RUN useradd -l -u $UID app -d /app
+RUN groupadd -g $UID app
+RUN useradd -l -u $UID app -g $UID -d /app
```
lauterman, line 9: it should be -g $GID

jrpatterson: I didn't define a GID variable so I just had it piggyback off of the GID value.  Would you add it ?

AndrewGearhart: Dropped this in for you... https://github.com/psu-libraries/researcher-metadata/pull/1019/commits/0cf0288ff308d28e74d7d4ee100333378782a1d2

ajkiessl (approved):
This lgtm.  @jrpatterson is this ready to go?  I'd like to merge it to main and pull it into the branch I'm working on.

PR #1028, opened 2024-12-03
Rescue Net::ReadTimeout when sending OA notifications.

### LINKED ISSUE #1018: Rescue and log `Net::ReadTimeout` when sending oa email notifications
In `OpenAccessNotifier#send_notification_to_user`, we rescue `Net::SMTPFatalError` and log it so the iterator continues to the next notification.  We have, however, run into other `Net` errors.  Specifically: `ReadTimeout`.   This could be caused by a temporary issue with the connection to the mail server, so I think it makes sense to rescue it as well, log it, and move to the next notification.

ajkiessl:
Fixes part of #1018 .  Still need to figure out what is causing the Net::ReadTimeout.  Hopefully this will help, though.

### DIFF
--- a/app/models/open_access_notifier.rb
+++ b/app/models/open_access_notifier.rb
@@ -40,7 +40,7 @@ class OpenAccessNotifier
             p.authorships.find_by(user: user).record_open_access_notification
           end
         end
-      rescue Net::SMTPFatalError => e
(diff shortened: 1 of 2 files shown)


jlandiseigsti (approved):
LGTM!

PR #1030, opened 2024-12-04
Allow get and delete methods when signing out

ajkiessl:


Smullz622 (approved):
looks good, was this needed because of the upgrade?

ajkiessl:
@Smullz622 It actually was like this before the upgrade.  Rails admin defaults to `:delete` and the rest of the app uses `:get`.  It's actually more complicated to get all of the sign out buttons to use either `:delete` or `:get` than just setting the route to accept both.  At some point during my devise upgrades it wouldn't let me use both.  Not exactly sure what was causing it.  I realized after deploying that we need both, and for some reason it works now.

PR #1034, opened 2024-12-04
set permissions flags after OAB check

### LINKED ISSUE #825: When OAB returns no embargo and/or no set statement, set the flag(s) to indicate they've been checked
Currently, in the AI OA workflow, if the query to OAB indicates there is no embargo and/or no set statement, the field(s) remain blank and the "confirmed" flag remains "false". We should change this so the fields remain blank but the "confirmed" flag gets set to "true". This will enable such records to bypass manual review of permissions metadata. 

Related issue from the Standard OA Workflow is #571.

Smullz622:
Fixes #825


ajkiessl (approved):
LGTM

PR #1035, opened 2024-12-05
oa email notifications- filter out pubs with an associated AI OA file

### LINKED ISSUE #1027: Do not send OA Notification emails out for publications with Activity Insight OA Files
In the `User#needs_open_access_notification` filter, let's remove the `activity_insight_postprint_status` `where` clause:

`.where('publications.activity_insight_postprint_status IS NULL OR publications.activity_insight_postprint_status != ?', 'In Progress')`

And instead, filter out any publication that has an associated Activity Insight OA File.

@anaelizabethenriquez If a user does receive an OA notification email (for a different publication that does not have an Activity Insight OA File) should publications with Activity Insight OA Files be filtered out of the email?  We don't currently filter out publications with a post print status of "In Progress" from these emails.

Smullz622:
Fixes #1027


ajkiessl (approved):
LGTM thanks!

PR #1039, opened 2024-12-06
add 'Flagged for Review' bucket to AI OA Workflow

### LINKED ISSUE #978: Create a new bucket in the AI OA Workflow to hold "problem" publications/files
In most of the buckets in the AI OA Workflow, there is a way for a publication to be "weird" and get stuck in the bucket. For example:

- Publications in the DOI review bucket that are not yet published: We anticipate they will get a DOI, but we don't know it yet. Or we anticipate that they won't, but we can't be sure until they are published.
- Publications where the file has tracked changes that need to be accepted or extra content that shouldn't be deposited in ScholarSphere (such as a letter from the journal editor) (issue #947 would allow us to resolve these in the workflow -- until then I am resolving them outside it)

These files end up stuck at the top of the list of publications/files to review until I (Ana) get a chance to clean them up. It would be great to have another bucket for these items and a Boolean to toggle so that they won't show up in any of the other buckets (except the troubleshooting bucket -- everything in the workflow should be there). This bucket could be the second-to-last one on the list, above the troubleshooting bucket.

Suggested wording for this bucket:
Flagged for Review
Publications that an admin user has flagged due to unusual issues with the publication or associated files

Smullz622:
Fixes #978 

Adds all publications with `flagged_for_review == true` to new bucket and removes them from all other buckets.

With the rails upgrade, running `db:migrate` made changes in the schema to all timestamped attributes. Default precision is now 6 so declaring that is not necessary and anything that's not 6 needs to specify precision nil.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,6 +49,7 @@ RSpec/LeakyConstantDeclaration:
 RSpec/MultipleMemoizedHelpers:
   Exclude:
     - 'spec/component/models/user_spec.rb'
+    - 'spec/component/models/publication_spec.rb'
 
 Style/Semicolon:
   Exclude:
--- a/app/components/activity_insight_oa_dashboard_component.html.erb
+++ b/app/components/activity_insight_oa_dashboard_component.html.erb
@@ -124,6 +124,24 @@
     </div>
   <% end %>
   <br>
+  <% if flagged_for_review_count.nonzero? %>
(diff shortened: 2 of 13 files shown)


ajkiessl (approved):
Looks good.  Just one suggestion below.

ajkiessl (approved):
LGTM

PR #1041, opened 2024-12-09
983 bugfix accepted version

### LINKED ISSUE #983: Accepted versions are getting labeled as published versions in AI OA Workflow
Lucretia has noticed quite a few files in the "Review Wrong File Versions" bucket where the "Version" field on the file is incorrect. As of now, these are all marked as "publishedVersion," but they're really the accepted version.

https://metadata.libraries.psu.edu/admin/activity_insight_oa_file/2049
https://metadata.libraries.psu.edu/admin/activity_insight_oa_file/1838
https://metadata.libraries.psu.edu/admin/activity_insight_oa_file/1837
https://metadata.libraries.psu.edu/admin/activity_insight_oa_file/2217
https://metadata.libraries.psu.edu/admin/activity_insight_oa_file/791

These files have a lot in common. They are all PDFs created with LaTeX, and they have the arXiv watermark on them. I suspect there is something in the automated version checker that is mistakenly categorizing these as published. Guessing this will require some development time to fix and needs to wait until fall, but giving @ajkiessl a heads up just in case.

Ideally I'd like to capture whatever we'll need for development in this issue so we can proceed with correcting this error manually for the files we have now.

jlandiseigsti:
closes #983


app/models/file_version_checker.rb
```diff
@@ -35,8 +35,11 @@ def process_content
 
       begin
         reader = PDF::Reader.new(file_path)
+        if contains_arxiv_watermark?(reader)
+          @score = -9999
+          return ''
+        end
```
ajkiessl, line 41: I think it would probably be good to pull `reader = PDF::Reader.new(file_path)` out of this and memoize it in its own function: 
```
def reader
   @reader ||= PDF::Reader.new(file_path)
end
```

Then, move `if contains_arxiv_watermark?(reader)` up to the `version` method, and have it return accepted version.  Forcing the score down to -9999 feels like a work around that I don't think is necessary, but I don't know exactly how the PDF::Reader object operates.  I could be wrong if it operates like an enumerator.

ajkiessl (changes requested):
Looking good.  One request below.

ajkiessl (approved):
LGTM

PR #1042, opened 2024-12-09
Sort by doi in 'Publications Requiring Preferred Version Check' table

### LINKED ISSUE #965: On Preferred Version page, include DOI in metadata display and sort by DOI
This page: https://metadata.libraries.psu.edu/activity_insight_oa_workflow/preferred_version_review

Sorting by DOI would enable us to review all of the publications covered by a single publisher's policy at one time. 

Currently, the data on this page is (IIRC) sorted by the date the file entered RMD. Instead, we should sort by the DOI of the associated publication. That DOI should also be displayed, in a new column. The user does not need to be able to change the sorting or sort on other fields during use.

ajkiessl:
closes #965


Smullz622 (approved):
LGTM!

PR #1043, opened 2024-12-10
ensure doi matching is not case sensitive

### LINKED ISSUE #936: DOI matching should not be case sensitive
Looking at some publications that either weren't grouped into duplicate groups or weren't merged, I think we need to check all our DOI-matching logic to ensure it's case insensitive (as I believe URLs and DOIs are).

This change should happen throughout the RMD application, in the duplicate grouping/merging scripts, in the DOI confirmation process, in the importers, and anywhere else we match on DOI.

Smullz622:
Fixes #936 

Grouping appears to be the only place that had case sensitive doi matching. The auto merge matching and importers already  handle casing.


jlandiseigsti (approved):
LGTM!

ajkiessl (approved):
Looks good, thanks!

PR #1044, opened 2024-12-10
973 doi merge fix

### LINKED ISSUE #973: DOI disappears on merge
I've been cleaning up a user's publications this morning and using the feature to group duplicates, then reviewing and merging them. A couple of times, I have selected the record with a DOI as the target for the merge, but upon merge, the DOI is gone. I thought the DOI was supposed to stay no matter what, and that in the case of two DOIs, it would use the DOI for the target publication. Let's check on this.

A little more on this: I've seen this happen several times with DOIs that have "True" as the value in "DOI verified?" It just happened with this DOI: https://doi.org/10.52842/conf.ecaade.2016.1.501 I wondered if it could be a validation issue, but that DOI looks valid to me, and I am able to add it back to the record manually.

EricDurante:
fixes #973 

These unit tests are bonkers, but that's because the logic behind this behavior is kind of bonkers, and I want these tests to be thorough since ensuring the correct behavior in every circumstance here is crucial. We could separate the persistence logic from the business logic here, but that wouldn't simplify much. I suppose that we could extract one or more other policy objects from this class, but I'm not sure what we would gain from that at this point. It all reads well enough to me, and I think I'm happy with it as-is for now unless anybody has any suggestions.

### DIFF
--- a/app/controllers/admin/publication_merges_controller.rb
+++ b/app/controllers/admin/publication_merges_controller.rb
@@ -41,6 +41,10 @@ class Admin::PublicationMergesController < RailsAdmin::ApplicationController
                                           merge_target_publication_id: params[:merge_target_publication_id],
                                           selected_publication_ids: params[:selected_publication_ids])
         return
+      rescue DOIVerificationMergePolicy::UnmergablePublications
+        flash[:error] = I18n.t('admin.publication_merges.create.unmergable_publications_error')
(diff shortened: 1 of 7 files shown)


EricDurante:
Not sure why there's a failing test. It seems to be unrelated to the changes here and only failing intermittently.

PR #1045, opened 2024-12-11
exclude extension publication pub types from OAB & unpaywall 

### LINKED ISSUE #962: Exclude the publication type "Extension Publication" from being updated by OA Button and Unpaywall
I thought we had an issue for this, but I can't find it.

@nmg110 and I have heard several times now from a developer in the College of Ag who is using the RMD API. [Publication 538574](https://metadata.libraries.psu.edu/admin/publication/538574) is an Extension Publication ([Vaccination for Highly Pathogenic Avian Influenza (HPAI)](https://extension.psu.edu/vaccination-for-highly-pathogenic-avian-influenza-hpai)) that does not have a DOI. The DOI field should be blank. Instead, one of the importers keeps filling in an incorrect DOI, for a journal article that has a similar title: "Rational approach to vaccination against highly pathogenic avian influenza in Nigeria: a scientific perspective and global best practice" from Archives of Virology (doi: https://doi.org/10.1007/s00705-023-05888-2)

As @ajkiessl  shared in an email back in December, querying OA Button with the correct title leads to this record, which has the title for the PSU Extension publication but the DOI and other metadata for the Archives of Virology article: https://bg.api.oa.works/find?id=Vaccination%20for%20Highly%20Pathogenic%20Avian%20Influenza

I have reported this error to OA Button but not heard back. Is there anything we can do on the Penn State side to stop this happening? I thought about changing the "authorships" in RMD on this publication to "unconfirmed," but the authors do want the article to show up in their data. How difficult/how much of a bad idea would it be to exclude the publication type "Extension Publication" from being updated by OA Button and Unpaywall? I don't think any of them (at least published by PSU) have DOIs (at the moment; could change in the future). Any other ideas?

Smullz622:
Fixes #962 

I was debating about where to put the logic for this. Code wise the most efficient approach is putting it in the unpaywall/OAB clients so we're never actually making a call out on extension publications & just returning an empty hash as the response (similar to what we would get if the publication couldn't be found) as opposed to adding logic to everywhere we make unpaywall/OAB calls. However, in many of those places we're processing that response  (like `Publication#update_from_unpaywall`) & it seems like it would be functionally more efficient to put `return if pub type == extension publication` at the beginning of the method and skip all of that processing. I went with putting it in the client to keep it DRY, but I'm open to thoughts/suggestions.

This also has some code clean up. The rake task is no longer used & calls an `UnpaywallPublicationImporter` method that no longer exists. The spec was using a different method than the describe block is testing.

### DIFF
--- a/app/importers/oab_client.rb
+++ b/app/importers/oab_client.rb
@@ -8,7 +8,7 @@ class OABClient
                  "https://api.openaccessbutton.org/find?title=#{CGI.escape(cleaned_title(publication))}"
                end
 
-    json = JSON.parse(HttpService.get(find_url))
+    json = publication.publication_type == 'Extension Publication' ? {} : JSON.parse(HttpService.get(find_url))
     OABResponse.new(json)
   end
 
--- a/app/importers/unpaywall_client.rb
+++ b/app/importers/unpaywall_client.rb
@@ -6,9 +6,11 @@ class UnpaywallClient
       doi_url_path = Addressable::URI.encode(publication.doi_url_path)
       find_url = "https://api.unpaywall.org/v2/#{doi_url_path}?email=openaccess@psu.edu"
       json = JSON.parse(HttpService.get(find_url))
-    else
+    elsif publication.publication_type != 'Extension Publication'
       find_url = "https://api.unpaywall.org/v2/search/?query=#{CGI.escape(publication.title)}&email=openaccess@psu.edu"
       json = JSON.parse(HttpService.get(find_url))['results'].blank? ? '' : JSON.parse(HttpService.get(find_url))['results'].first['response']
+    else
+      json = {}
     end
 
     UnpaywallResponse.new(json)
--- a/lib/tasks/imports.rake
+++ b/lib/tasks/imports.rake
@@ -136,33 +136,6 @@ namespace :import do
     PurePublicationImporter.new.call
     PurePublicationTagImporter.new.call
   end
-
-  desc 'Import all data'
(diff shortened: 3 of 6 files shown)


EricDurante:
I can see why you were debating. I actually have a strong preference here. I'd argue that the API client objects should have one job, which is to communicate with their respective services. I wouldn't want to mix application business logic of any kind (including publication filtering) into those classes because it will tend to make them far less flexible and reusable. However, I absolutely agree that we should try to keep this logic in one place.

I think that it would be better if we never even load this type of publication out of the database if we don't intend to do anything with it. So rather than putting repetitive guard clauses in places like `Publication#update_from_unpaywall`, I think that we should probably update the database queries that load the publications to be checked in the first place. For instance, here: https://github.com/psu-libraries/researcher-metadata/blob/aa633b3d3ba805a2d33ad1f9e8d33ce9499f0c88/app/importers/unpaywall_publication_importer.rb#L7
If we need this same logic elsewhere, we could put it into a scope on the `Publication` model. That way, things stay fairly DRY, and then we're not even loading or iterating over irrelevant publication records, much less making external API calls for them.

Does anyone else have thoughts/preferences?

Smullz622:
I've moved the logic out of the clients to where the publications are being loaded in the first place. Side note- there are OAB and Unpaywall calls that are not covered by this, but they are all within the activity insight oa workflow. At this time, Extension Publications are not oa publication types and therefore already excluded from the workflow.

ajkiessl (approved):

PR #1046, opened 2024-12-11
Exclude publications in PSU law school repos from AI OA workflow

### LINKED ISSUE #935: If the publication is in one of the law school repositories, do not put it through the AI OA Workflow
If RMD already has an OA location with a source of Dickinson Ideas or Penn State Law eLibrary, the publication should be treated the way it would be treated if it had an OA location with a source of ScholarSphere. Rather than sending the file through the AI OA Workflow and depositing it in ScholarSphere, RMD should write to Activity Insight that the file is already openly available.

EricDurante:
fixes #935

I think that's all that there is to this issue. Am I missing anything?


Smullz622 (approved):
LGTM!

PR #1053, opened 2024-12-16
735 limit oa emails

### LINKED ISSUE #735: Cap the number and age of articles mentioned in OA reminder emails
Some RMD users have many articles that are subject to the OA workflow and aren't yet OA. When our reminder emails mention more than a few articles, this is often discouraging to the user. Given the responses we receive to these emails, we suspect many such users are not responding, just getting waivers or ignoring the email.

This has always been a problem for users with very high publishing output, but it's becoming a problem for more users as time passes (since we include any article with a publication date after July 1, 2020).

The feature request is:
* If the article was published more than two years ago, don't mention it in reminder emails or allow it to trigger reminder emails.
* Cap the total number of articles that a reminder email can mention, so that only the 6 articles with the most recent publication date get mentioned. (I don't think we have an easy way to correlate the number of publications mentioned in an email with the percent of users who take action based on it. If we did, we could use that data to pick this number, but since we don't, I think we can just guess.)

Within RMD, the user should still be able to see that these older articles are not openly available and be able to get a waiver, add a link, or deposit to ScholarSphere.

In #593, the new text will eliminate the distinction between "new" and "previous" publications.

jlandiseigsti:
closes #735


lib/mailer_previews/faculty_notifications_mailer_preview.rb
```diff
@@ -11,7 +11,7 @@ def open_access_reminder
     old_fake_pubs = [pub1, pub2]
     new_fake_pubs = [pub3]
 
-    FacultyNotificationsMailer.open_access_reminder(fake_user, old_fake_pubs, new_fake_pubs)
+    FacultyNotificationsMailer.open_access_reminder(fake_user, old_fake_pubs + new_fake_pubs)
```
EricDurante, line 14: Minor thing, but this could be cleaned up a little bit to get rid of the local assignments above and the array concatenation.

i.e.
```ruby
FacultyNotificationsMailer.open_access_reminder(fake_user, [pub1, pub2, pub3])
```

lib/tasks/email_notifications.rake
```diff
@@ -53,8 +53,7 @@ namespace :email_notifications do
     old_fake_pubs = [pub1, pub2]
     new_fake_pubs = [pub3]
     FacultyNotificationsMailer.open_access_reminder(test_user,
-                                                    old_fake_pubs,
-                                                    new_fake_pubs).deliver_now
+                                                    old_fake_pubs + new_fake_pubs).deliver_now
```
EricDurante, line 56: Same thing here as above.

EricDurante (changes requested):
I commented on a couple of individual instances, but the same applies to each of the spec files also. Since the mailer's interface changed to accept one array of publications instead of two, I think that it's worth updating those spots in the tests and other places where we're now passing the mailer a concatenation of the two former array arguments.

The separation of those two groups of publications is now a meaningless vestige of the way that things used to be. As a developer, if I didn't know the history, I would wonder why those arrays were separated to begin with since they're just being concatenated anyway. I would think that I was missing something. So I'd argue that it's both needless extra code and also needless potential confusion for future developers to leave things that way. I'd like to see each of those separate arrays just combined into one and passed as one to the mailer.

Other than that, this looks good!

EricDurante (approved):

ajkiessl (approved):

PR #1054, opened 2024-12-16
 export publications by user

### LINKED ISSUE #426: Export publications by user
I would like to be able to go to the Publications section, search for a specific user and then export those publications to a CSV. Right now I have to pull the information via the API to JSON, then clean and format the data so I can then import it to excel. Once in excel I then have to clean and format the data again for it to be imported to AI. This might be a mute point if we complete issue #425 .

Smullz622:
Fixes #426

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -544,7 +544,7 @@ class Publication < ApplicationRecord
     end
 
     scope do
-      Publication.joins(:users)
+      Publication.left_joins(:users)
     end
   end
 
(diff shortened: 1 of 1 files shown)


ajkiessl (approved):
LGTM

app/models/publication.rb
```diff
@@ -536,6 +542,10 @@ def doi_url_path
         field(:preferred_file_version_none_email_sent)
       end
     end
+
+    scope do
```
EricDurante, line 546: Looks good. Did the omission of this scope cause a test to fail?

Smullz622: Omission actually caused the export process to fail. The tests failed because I originally had `joins` instead of `left_joins`

EricDurante (approved):

PR #1055, opened 2024-12-18
Create an endpoint for ScholarSphere webhooks to delete open access locations in RMD

### LINKED ISSUE #766: When a work is withdrawn or deleted in ScholarSphere, remove the Open Access Link in RMD
Edited to include ScholarSphere's options for "withdrawn" and "deleted" works:

When a work is withdrawn or deleted in ScholarSphere, remove the Open Access Link in RMD. We'd like to remove Open Access Links with the source "ScholarSphere" if the work is withdrawn or deleted in ScholarSphere. This should apply to works deposited using the RMD OA workflow as well as works deposited directly in ScholarSphere.

There are also Open Access Links pointing to ScholarSphere that have OA Button or Unpaywall as the source. I don't think we don't need to remove these. OAB and Unpaywall should be removing them from their databases automatically, and the regular imports should pick that up. Right?

EricDurante:
closes #766 

I think I made this about as simple as I could.

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class WebhooksController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  before_action :authenticate_request
+
+  def scholarsphere_events
+    if params[:publication_url].present?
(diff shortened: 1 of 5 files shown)


ajkiessl (approved):
This LGTM.  I will add the necessary secrets to vault .

Issue #1037, opened 2024-12-05
Periodically, run the DOI verification process on existing publications


anaelizabethenriquez:
We should set up the DOI verification process to run periodically (perhaps once per month, depending on how long it takes to run) on existing publications for which the DOI is not confirmed (i.e., the DOI confirmed field is either "false" or "null"). This is related to #1036 , which is about running it on new publications as they come into RMD.

An extra motivation to do this is that we know we've been losing DOIs on merge for some publications. (see #973)

When this goes live, we'll want to make sure this process has a chance to run to completion on the full set of publications and then automated deduplication can occur and the Unpaywall and OAButton imports can run. We may want to pause reminder emails for a week or two related to this.

Update: The impact of this is smaller than I thought it would be. See #1036. Still working on rewriting/reprioritizing this.

anaelizabethenriquez:
For the first run of this, we want to run it on all publications (except those with type "Extension Publication"), regardless of whether the DOI is confirmed. This should be one-time only.

anaelizabethenriquez:
Publications that have a doi_error should go through the periodic checks.

@jlandiseigsti Is it possible for an item to get stuck in an oa_workflow_state of 'automatic DOI verification pending'? Assuming that publications are only in that state briefly when DOI verification is actually happening, I would say the periodic checks should skip them.

anaelizabethenriquez:
Sounds good. I see that quite a few of the pubs with a DOI error are stuck in that status. Thanks, @jlandiseigsti !

EricDurante:
Sorry that I'm late to the conversation, but just to add what I know:

A `Publication`'s `oa_workflow_state` isn't actually an indicator of it's current "position" in the workflow at any given time. It's more just a means of excluding `Publication`s from scopes where they don't currently belong. So when we've queued up a job to verify the DOI for a`Publication`, we set the `oa_workflow_state` to `"automatic DOI verification pending"` just as a way of excluding that `Publication` the next time we ask for `Publication.needs_doi_verification` to queue up the next batch of DOI checks. In other words, it's just our way of saying, "this publication that's missing DOI information is already queued up to have the DOI checked, so don't queue it up again".

That's why we don't bother clearing `oa_workflow_state` after the check is complete. After that information is used to appropriately exclude something from a given scope, then it has done its job, and we don't check it again or use it for any other purpose until it's time to set that field to some other value in order to exclude the `Publication` from some other scope later in the workflow.

[...]

PR #1060, opened 2025-01-08
1036 doi verification

jlandiseigsti:


app/importers/activity_insight_importer.rb
```diff
@@ -227,7 +228,6 @@ def call
               unless pub_record.doi_verified == true
                 pub_record.oa_workflow_state = 'automatic DOI verification pending'
                 pub_record.save!
```
jlandiseigsti, line 230: There's one edge case that I'm waiting to hear back from Ana about. Before, if the AI OA File had a different location as the imported one, it would run the DOI verification, even if the publication wasn't new. I removed the DOI verification from line 230 because I don't want double verification for the ai oa imports files that are new but I'm waiting to confirm that it's ok to lose the verification on the others (since we're planning to implement routine DOI checks).

jlandiseigsti: Update: Ana said it's fine to remove this part since we're doing checking with both this ticket and #1037

app/importers/activity_insight_importer.rb
```diff
@@ -166,7 +166,12 @@ def call
           if pi.persisted?
             update_pub_record(pub_record, pub)
           else
-            pi.save!
+            unless pub_record.doi_verified == true
+              pub_record.oa_workflow_state = 'automatic DOI verification pending'
+              pub_record.save!
+              DOIVerificationJob.perform_later(pub_record.id)
+              pi.save!
+            end
```
Smullz622, line 174: This is also going to save a workflow state for things that don't belong in the workflow. If you want it to be here, you could put a conditional around setting the state here. Alternate could be to leave that logic where it was & have `unless pub_record.doi_verified == true` be `unless pub_record.doi_verified == true || pub_record.oa_workflow_state == 'automatic DOI verification pending' to avoid queueing the job multiple times

Smullz622 (changes requested):

app/importers/activity_insight_importer.rb
Smullz622: I like the change of using `#update` here. I think `#can_receive_new_ai_oa_files?` covers most of what we want, but it doesn't account for whether there is an activity insight file location. Easy fix could be to just move `activity_insight_file_location = pub.postprints&.first&.location` up & add `activity_insight_file_location.present?` to the current conditional. We have a a publication scope `activity_insight_oa_publication` that would cover all conditions, but I'm not sure if that would be useful since we would want to check `pub` for a file location and `pub_record` for conditions covered by `#can_receive_new_ai_oa_files?`

Smullz622 (approved):
lgtm!

PR #1061, opened 2025-01-08
create bucket for wrong file version author notified 

### LINKED ISSUE #997: Updates to Wrong File Version bucket
Currently, 611 of 1196 publications in the AI OA Workflow are in the "Review Wrong File Version" bucket. It's turning out to be quite difficult to get publications out of this bucket. The response rate to our emails is low. There are more than 400 publications in the bucket that we've sent emails for already. #996 is a feature aimed at that problem. But we know from other emails we send (OA reminder emails, ScholarSphere curation emails) that the response rate is not going to be 100%.

We need to do something with publications for which we send repeated emails and the researcher does not reply. 

### Our plan

- [x] Add a count of how many times we've sent a Wrong Version email about the file. (This count should be on the file record, not the publication record. If the file gets replaced in Activity Insight, the new file should have a count of 0 when imported into RMD.) Display this in the tables for the Wrong Version bucket and a new "Wrong Version - Author Notified" bucket.
- [x] Create a new bucket, "Wrong Version - Author Notified". Files belong in this bucket if we have already sent one or more wrong version emails about the file and the associated publication does not have an unknown file version or correct file version. Description for list of buckets: "Publications that do not have an unknown file version or correct version, for which the author has already been notified." The table in this bucket should have the files grouped by author and have the same columns as the existing "Wrong Version" bucket. This author groups should be sorted so authors emailed most recently are at the top of the list. There should be a button to send a Wrong Version email about an individual file and a button to send a batch email about all the files for the author in this bucket. (Files in the other Wrong Version bucket should not be mentioned.)
- [x] Adjust the current Wrong Version bucket so it does not include files for which we have already emailed the user. Adjust the batch emails sent from this bucket to ensure they only mention files that are in this bucket.

Smullz622:
Fixes #997 

Creates bucket for AI OA Workflow publications with only wrong file versions where authors have already been notified about each file associated with the publication.


app/components/activity_insight_oa_dashboard_component.html.erb
```diff
@@ -70,6 +70,24 @@
     </div>
   <% end %>
   <br>
+  <% if wrong_version_author_notified_count.nonzero? %>
+    <%= link_to activity_insight_oa_workflow_wrong_version_author_notified_review_path, class: 'card-link' do %>
+      <div class='card' id='wrong-version-author-notified-check-card'>
```
jlandiseigsti, line 75: Long term thought – it would make this page more concise if all these cards, with their header and body, were another component that could be given active?, header_text, and body_text properties. But probably not worth refactoring atm.

app/models/publication.rb
```diff
@@ -220,7 +220,7 @@ def self.preferred_version_options
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND publications.preferred_version = activity_insight_oa_files.version)})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version IS NULL)})
   }
-  scope :wrong_file_version, -> {
+  scope :wrong_file_version_base, -> {
```
jlandiseigsti, line 223: 👍

app/models/publication.rb
```diff
@@ -241,6 +241,14 @@ def self.preferred_version_options
...
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version IS NULL)})
   }
+  scope :wrong_file_version, -> {
+    wrong_file_version_base
+      .where(%{EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND (activity_insight_oa_files.wrong_version_emails_sent IS NULL OR activity_insight_oa_files.wrong_version_emails_sent = 0))})
+  }
+  scope :wrong_version_author_notified, -> {
+    wrong_file_version_base
+      .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND (activity_insight_oa_files.wrong_version_emails_sent IS NULL OR activity_insight_oa_files.wrong_version_emails_sent = 0))})
+  }
```
jlandiseigsti, line 251: Might be able to use the `invert_where` method to avoid repeating this SQL logic. 

https://edgeapi.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-invert_where

spec/component/models/publication_spec.rb
```diff
@@ -1179,6 +1189,12 @@
       end
     end
 
+    describe '.wrong_version_author_notified' do
+      it "returns activity_insight_oa_publications whose associated files' versions does not contain an 'unknown' version or correct version" do
```
jlandiseigsti, line 1193: Might be worth mentioning in the description that these are also ones where a least one email has been sent

app/views/activity_insight_oa_workflow/wrong_version_author_notified_curation/index.html.erb
```diff
@@ -0,0 +1,67 @@
+<div class="container pt-4 pb-5">
```
jlandiseigsti, line 1: Similarly to the dashboard cards, seems like the actual pages could use some refactoring to avoid duplication, since this page is very similar to the `wrong_version_curation` page (and others). But also similarly, not sure it's worth the time commitment.

jlandiseigsti (changes requested):
I'd like the test I commented on to have a little more descriptive of a name, but for the de-duplication refactors, I leave up to your judgment.

jlandiseigsti (approved):
Those dashboard cash component changes really make that whole page so much more readable...and will be great in the event on any changes too. Thanks for taking the time to refactor!

app/models/publication.rb
EricDurante: Agreed that it would be nice to eliminate the duplication here. Not so sure about `invert_where` since it could have unintended consequences. I think my preference would be to just define the common inner select statement as a string literal somewhere and then interpolate it into the outer query in both of the `where` calls.

Also, I'm guessing that we don't need to bring back all of the fields in those inner `SELECT`s since we're just checking for the existence of records. So, `SELECT id` or something instead of `SELECT *`?

app/views/activity_insight_oa_workflow/wrong_version_author_notified_curation/index.html.erb
EricDurante: In my experience, refactoring duplicated code almost always provides a big return on investment. I'd vote for it.

spec/component/models/publication_spec.rb
EricDurante, line 1: There are actually two new scopes, `.wrong_version_author_notified` and `.wrong_file_version_base`. I'd add a test for `.wrong_file_version_base` as well. Even though it's currently only used by other scopes in `Publication`, it's in the public interface and available to be used anywhere.

The better thing to do instead of adding a test for a public method that's not actually used publicly would be to make `.wrong_file_version_base` private, but I'm not aware of any nice way to do that. We could probably define `.wrong_file_version_base` as a regular class method instead of using the `scope` syntax and make it private, but that's kind of ugly. We could extract the whole business into a "query object" instead of using ActiveRecord scopes and then make the internals private, but that feels a bit cumbersome.

In short, all of the patterns that I'm familiar with kind of suck more than just leaving `.wrong_file_version_base` public and adding a test for it.

EricDurante (changes requested):

jlandiseigsti (approved):
Looks good to me! Thanks for making those refactors

EricDurante (approved):
Looks good!

ajkiessl (approved):

PR #1062, opened 2025-01-13
updates to permissions metadata

### LINKED ISSUE #1052: Additional permissions metadata improvements
Related to #825, a couple things I failed to describe in that issue:

- [x] If OAButton does not return a license, we should set that field to "All Rights Reserved."

- [x] We should do a one-time update to the existing records once the change here and the changes in #825 are in place, so that the flags get set appropriately for items currently in the workflow.

Smullz622:
Fixes #1052 

Set license to 'All rights reserved' when OAButton returns permissions, but fails to return a license. Applies this logic to `checked_for_set_statement` and `checked_for_embargo_date` flags as well - flags are currently set to true whenever OAB is checked, but they should only be set when OAB returns permissions without set statement/embargo date.

Also has one time rake task that will run all AI OA files through new permissions check (flags when no set statement or embargo found & sets license)

### DIFF
--- a/lib/tasks/oa_workflow.rake
+++ b/lib/tasks/oa_workflow.rake
@@ -4,3 +4,11 @@ desc 'Move publications through the open access workflow'
 task oa_workflow: :environment do
   OAWorkflowService.new.workflow
 end
+
+desc 'One time update to set all permissions flags'
+task permissions_check_all: :environment do
+  ActivityInsightOAFile.all.each do |file|
+    file.update!(permissions_last_checked_at: Time.current)
+    FilePermissionsCheckJob.perform_later(file.id)
+  end
+end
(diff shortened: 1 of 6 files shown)


jlandiseigsti (approved):
Looks good to me!

lib/tasks/oa_workflow.rake
EricDurante, line 1: I think this is fine, but since this is queueing all of these checks up as a series of async background jobs, I guess we'll just want to make sure that we verify that each job has actually run and succeeded after we've executed this task.

Maybe that's a given - I just don't really have any visibility into the world of RMD production job queues.

EricDurante (approved):

lib/tasks/oa_workflow.rake
Smullz622: I double checked - we will have visibility on the jobs through argo workflows & the delayed job pod

ajkiessl (approved):

PR #1065, opened 2025-01-14
fix duplicate publication rows

### LINKED ISSUE #1064: Publication appears more than once in search results
A couple of times today, I have noticed that the Publications section of the admin interface returns more results than I expect for a given search. I just realized why -- the same publication is appearing more than once. I'm attaching two screenshots. In each one, the publication ID is the same for all of the entries, but it's reported as "2 publications" or "3 publications" and shows the same pub on more than one line.

Image
Image

Tagging @Smullz622 since I would normally tag Alex when reporting a bug of this sort.

Smullz622:
Fixes #1064

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -545,7 +545,7 @@ class Publication < ApplicationRecord
     end
 
     scope do
-      Publication.left_joins(:users)
+      Publication.left_joins(:users).distinct
     end
   end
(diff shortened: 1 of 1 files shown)


jlandiseigsti (approved):
Took me a while to get through this PR, but it looks good, unless I missed something

PR #1073, opened 2025-01-22
adds images and file download link to wrong version email

### LINKED ISSUE #996: Link to file and add help text to Wrong File Version email
Often, when a researcher responds to the "Wrong File Version" email, they send us the same file that we already have. It would be nice to attach the file(s) we have to the email, or at least include their file names. Update: rather than attaching the files, given Eric's comments below, we're going to link to them so the user can download them.

We would also like to add the example images illustrating the different versions directly to the article. We link to [this FAQ](https://psu.libanswers.com/faq/273465) that includes them, but researchers aren't getting the message.

We should make the following changes to the wrong file version email:

- In the table listing wrong files, add a link to view/download each file that we currently have.
- After the sentence "Some journals prefer that you share the final published version." add the heading "Examples" and then two images, one labeled "accepted manuscript" and one labeled "final published version". These can be copied from the FAQ linked above and should have the alt text used there (unless you have improvements to the alt text, in which case please let Ana know and she'll fix it in the FAQ). There are four images in that FAQ -- for this email, we want to use these two:
-- [AM2.png](https://d2jv02qf7xgjwx.cloudfront.net/accounts/182707/images/AM2.png)
-- [VOR2.png](https://d2jv02qf7xgjwx.cloudfront.net/accounts/182707/images/VOR2.png)

Smullz622:
Fixes #996


jlandiseigsti (approved):
Looks good to me!

ajkiessl (approved):
LGTM

PR #1074, opened 2025-01-22
1069 permissions review doi

### LINKED ISSUE #1069: On Permissions Review page, include DOI in metadata display and sort by DOI
Same as #965 but for a different page in the workflow

This page: https://metadata.libraries.psu.edu/activity_insight_oa_workflow/permissions_review

Sorting by DOI would enable us to review all of the publications covered by a single publisher's policy at one time.

Currently, the data on this page is (IIRC) sorted by the date the file entered RMD. Instead, we should sort by the DOI of the associated publication. That DOI should also be displayed, in a new column. The user does not need to be able to change the sorting or sort on other fields during use.

jlandiseigsti:
closes #1069


EricDurante (approved):
This looks good to me

PR #1077, opened 2025-01-23
Fix bug in publication DOI merge policy

### LINKED ISSUE #1076: Publications in a duplicate group will not merge if the DOIs are verified and vary by case
This is the error text: "You have attempted to merge two or more publications that cannot be merged because they have distinct, verified DOIs. Please double-check the publications in this group."

DOIs that vary only by case (uppercase/lowercase) should not be treated as distinct. I think this bug is coming up because of the expansion of the DOI verification logic that we rolled out earlier in this development round.

EricDurante:
Fixes #1076


jlandiseigsti (approved):
LGTM!

PR #1078, opened 2025-01-24
Remove surrounding HTML paragraph tags from publication abstracts

### LINKED ISSUE #939: Remove paragraph tags from beginning and end of abstract field
Many publication abstracts in RMD begin with `<p>` and end with `</p>`. Users have to remove these manually before using that metadata for a ScholarSphere deposit. It would be great to remove those tags systematically, and also set up the relevant importer(s) to remove them upon import in the future. Other tags in the abstract field should probably remain as they are.

EricDurante:
Fixes #939

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -89,6 +89,8 @@ class Publication < ApplicationRecord
     ].freeze
   end
 
+  before_validation :strip_p_tags_from_abstract
+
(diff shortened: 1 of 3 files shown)


Smullz622 (approved):
lgtm!

ajkiessl (approved):

PR #1080, opened 2025-01-24
970 reminder email journal

### LINKED ISSUE #970: Include journal name and additional info in reminder emails
A relatively common response to reminder emails is "I'm not sure which article you are referring to. Could you give me the full citation?" In the publication list in the reminder email, let's include the citation that would appear for the article on the user's demo profile.

We'd probably want to implement this at the same time as #735 since this will add a lot of info to emails that can already be overwhelming. [Update: #735 is in place.]

jlandiseigsti:
closes #970


ajkiessl (approved):
LGTM

Issue #1047, opened 2024-12-11
Fix flaky test/s


ajkiessl:
This one seems to fail regularly:

spec/integration/admin/activity_insight_oa_workflow/wrong_file_version_review_spec.rb:71

I think there's another that similarly fails on the email sending tests in the AI OA workflow review lists.  Since this issue started to arise after updating Rails to 7.2, I suspect it has something to do with these clauses: `click_button('Send Email', match: :first)`.  I had to use this method of clicking the button since what I was using before became deprecated.

ajkiessl:
This one is also flaky in CI: spec/integration/admin/activity_insight_oa_workflow/file_version_review_spec.rb:50

This one is fixed: spec/integration/admin/activity_insight_oa_workflow/wrong_file_version_review_spec.rb:71

granta03:
Keeping issue open until we can confirm the PR #1091 is more or less flaky

ajkiessl:
It flaked: https://app.circleci.com/pipelines/github/psu-libraries/researcher-metadata/2239/workflows/06cc9745-208f-4fae-945b-84b8532c0b79/jobs/5918?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

PR #1091, opened 2025-01-27
Run jobs synchronously in test env with delayed job

ajkiessl:


granta03:
Made a change in spec/integration/admin/activity-insight-oa-workflow/file_version_review_spec.rb for the flaky test. If the test is less flaky we'll come back and fix the other similar flaky tests. Issue #1047

PR #1097, opened 2025-01-29
exclude files with inactive depositor from AI OA workflow

### LINKED ISSUE #920: Exclude files from AI OA workflow if user who uploaded to AI is no longer at Penn State
I believe this would prevent our spending time on deposits that end up failing because the depositor does not exist. E.g. https://metadata.libraries.psu.edu/admin/scholarsphere_work_deposit/1848

If the AI user associated with the file (the "uploader" or the "depositor") has a status of "MEMBER" (rather than staff or faculty, I think) in PSU Identity, exclude the file from the workflow. 

If the file has been uploaded by another user, the fact that a coauthor on the publication has the status of "MEMBER" should not exclude it from the workflow.

Smullz622:
Fixes #920

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -180,8 +180,10 @@ class Publication < ApplicationRecord
   scope :activity_insight_oa_publication, -> {
                                             oa_publication.with_no_scholarsphere_oa_locations
                                               .joins(:activity_insight_oa_files)
+                                              .joins('INNER JOIN users ON users.id = activity_insight_oa_files.user_id')
                                               .where.not(activity_insight_oa_files: { location: nil })
                                               .where('preferred_file_version_none_email_sent != true OR preferred_file_version_none_email_sent IS NULL')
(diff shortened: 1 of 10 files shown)


app/models/publication.rb
```diff
@@ -180,8 +180,10 @@ def self.preferred_version_options
   scope :activity_insight_oa_publication, -> {
                                             oa_publication.with_no_scholarsphere_oa_locations
                                               .joins(:activity_insight_oa_files)
+                                              .joins('INNER JOIN users ON users.id = activity_insight_oa_files.user_id')
                                               .where.not(activity_insight_oa_files: { location: nil })
                                               .where('preferred_file_version_none_email_sent != true OR preferred_file_version_none_email_sent IS NULL')
+                                              .where.not("users.psu_identity->'data'->>'affiliation' = ?", '["MEMBER"]')
```
EricDurante, line 186: Do we also need to have a similar condition in `ActivityInsightOAFile.subject_to_ai_oa_workflow`?

ajkiessl (approved):
LGTM

PR #1098, opened 2025-01-29
1090 cache source publication lists

### LINKED ISSUE #1090: Record the current list of publications (and their status) with each AI and Pure import
We need to be able to easily detect when a publication record that has previously been imported into RMD has subsequently been removed from the import source (Activity Insight or Pure) so that we can then decide whether to remove that publication record from RMD. In order to accomplish this, we don't want to (or simply cannot) request the publication metadata directly from the Activity Insight or Pure APIs for each record since RMD contains hundreds of thousands of publication records. Instead, each time that we import publications from Activity Insight and Pure, we should cache the current list of publications that are available from each import source in RMD's database so that we can compare those lists to RMD's publication records later. In the case of Activity Insight, we also want to record the status (published, in-press, etc.) of the publications in its list at the time of each import.

The lists of import source publications should be associated with a record of the import itself. The record of the import should capture when each import process began and when it completed so that we know what lists of import source publications to use for comparison.

This is prerequisite for #944.

EricDurante:
closes #1090

### DIFF
--- a/app/importers/activity_insight_importer.rb
+++ b/app/importers/activity_insight_importer.rb
@@ -6,6 +6,8 @@ class ActivityInsightImporter
   def call
     pbar = Utilities::ProgressBarTTY.create(title: 'Importing Activity Insight Data', total: ai_users.count)
 
+    import = Import.create!(source: 'Activity Insight', started_at: Time.current)
+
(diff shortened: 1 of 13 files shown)


ajkiessl (approved):
Looks good!

PR #1100, opened 2025-01-30
1050 update branding

### LINKED ISSUE #1050: Update public pages to match new PSU branding
This issue is for the pages that show for non-authenticated users, including https://metadata.libraries.psu.edu/ and a couple other pages linked from it.

jlandiseigsti:
closes #1050


ajkiessl (changes requested):
I think we should separate to public facing views and the non public facing views in `/layouts`, so we can just load the PSU assets for the few public pages, and keep the rest of the application the same as before.  So `application.html.erb` would be the same as before, but we'd add another layout `public.html.erb` for the `public/api_docs..home..resources.html.erb` views.

Loading the PSU assets in `application.htlm.erb`, is affecting parts of the application that we don't want it to.

PR #1120, opened 2025-02-17
1050 - changes the asset url to vary depending on env

jlandiseigsti:
There is a QA environment for the assets (including fonts) that we want our qa rmd site to point to. Making the external assets variable should address this.


ajkiessl (approved):
LGTM

PR #1149, opened 2025-06-13
Display book title in citation for Chapters instead of journal name

### LINKED ISSUE #1138: Display book title in citation for book chapters
The basic RMD "citation" format for book chapters does not include the title of the book in which the chapter appeared. We would like it to include the book title. 

Right now, I think our basic "citation" formatting is the same across all document types, but I think we would need to make it sensitive to document types in order to fix this. For journal articles, the Publisher field contains the name of the journal. For book chapters, it contains the name of the publisher. The Publisher field is included in the basic citation, so for journal articles the journal name appears, but for book chapter the much less relevant actual publisher appears.

An example is [Publication 416839](https://metadata.libraries.psu.edu/admin/publication/416839). In the user's profile, it displays as "[Transforming a Match Factory into the American Philatelic Research Library in Bellefonte, Pennsylvania](https://www.degruyterbrill.com/document/doi/10.1515/9783110679663/html), de Gruyter/Saur, 2021." RMD has the book title as the Secondary Title, but it does not display. Looking at a book chapter of my own, I see the same pattern. I'm not sure, though, if this data structure is consistent across all book chapters, imported from Pure and Activity Insight.

Related to #1059

ajkiessl:
closes #1138

### DIFF
--- a/app/decorators/authorship_decorator.rb
+++ b/app/decorators/authorship_decorator.rb
@@ -86,7 +86,7 @@ class AuthorshipDecorator < BaseDecorator
       # Workaround until we can address citations in depth
       # Chapters should reference the "secondary_title" instead of "published_by"
       # "secondary_title" generally stores the title of the book the chapter resides in
-      # This is preferred over "published_by" which is generally the publisher
+      # This is preferred over "published_by" which is the journal or publisher
       publication_type == 'Chapter' ? secondary_title : published_by
     end
 end


app/decorators/authorship_decorator.rb
```diff
@@ -77,8 +77,16 @@ def profile_management_pub_title
...
+      l += %{, <span class="journal-name">#{work_source}</span>} if work_source.present?
       l += ", #{year}" if year.present?
       l
     end
+
+    def work_source
+      # Workaround until we can address citations in depth
+      # Chapters should reference the "secondary_title" instead of "published_by"
+      # "secondary_title" generally stores the title of the book the chapter resides in
+      # This is preferred over "published_by" which is the journal or publisher
```
jlandiseigsti, line 89: Appreciate this comment.

jlandiseigsti (approved):
Looks good!

PR #1175, opened 2025-10-02
Add "deselect all" button to profile presentations page

### LINKED ISSUE #1136: Deselect all presentations in public profile
We have heard feedback from faculty, that they would like to be more selective with their presentations that display on their public profile. Because there are often so many presentations, we'd like to make it easy to deselect all presentations and then be able to select the ones that they do want to show.

granta03:
Fixes #1136


ajkiessl (approved):
LGTM

PR #1218, opened 2026-04-17
Fix /delayed_job - host not permitted error

### LINKED ISSUE #1159: /delayed_job gives "Host not permitted"
Investigate why we can't get to the delayed job interface at `/delayed_job`.

karthikreddy-02:
Fix /delayed_job "Host not permitted" error in production and QA

Summary
delayed_job_web uses Sinatra 4.x which introduced Rack::Protection::HostAuthorization as a hard-coded middleware that only allows localhost by default
Configured DelayedJobWeb with the app's permitted host per environment via :host_authorization setting in the initializer
Added request specs covering unauthenticated, non-admin, and admin access to the route
Closes #1159

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -3,6 +3,14 @@
 Delayed::Worker.max_attempts = 1
 Delayed::Worker.destroy_failed_jobs = false
 
+# Sinatra 4.x / rack-protection 4.x adds a strict HostAuthorization middleware that
+# by default only allows localhost. We must explicitly permit the app's host per
+# environment. Rails' ActionDispatch::HostAuthorization already protects the app at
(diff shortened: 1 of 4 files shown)


jlandiseigsti (approved):
Looks good! Appreciate the comments, and the tests are thorough as well.

One small note for next ticket (no need to change this one): we try to put the number of the ticket in the branch name, to make it easier to track down the reason why changes were made. So `delayed-job-host-not-permitted` would become `1159-delayed-job-host-not-permitted`, etc.

Let's be sure to test this in our QA environment, before deploying it. You probably know this already, but just as a reminder, after you merge, it will auto-deploy to ul-dev (this usually takes 10-15 minutes or so). You can then test it out there. Let me know if you need the RMD ul-dev link.

PR #1221, opened 2026-04-22
1199 doi verification job

### LINKED ISSUE #1199: `DOIVerificationJob` Unpaywall check failures
Image

This very likely from being rate limited by Unpaywall.  These send BugSnag on a ride, too, racking up over 80,000 errors in one day.  We need to handle errors here better.  I'm also not sure what's queuing thousands of `DOIVerificationJob`s.  That `HttpService` is designed to retry 10 times, so we can assume there's about 8,000 `DOIVerficationJob`s being triggered on a random Sunday.  Was it from a big import?

I have this error discarded in BugSnag.  Once this is fixed, we'll need to remember to undiscard it.

karthikreddy-02:
DOIVerificationJob had no error handling, so uncaught JSON::ParserError
exceptions from Unpaywall's non-JSON rate-limit (429) response bodies
bubbled out of the job and generated BugSnag alerts.

- Rescue StandardError in DOIVerificationJob#perform and log via
  ImporterErrorLog, matching the pattern in UnpaywallPublicationImporter
- Add `sleep 1 unless Rails.env.test?` throttle to ease API pressure
- Fix duplicate HttpService.get call in UnpaywallClient#query_unpaywall
  (search path was hitting Unpaywall twice per lookup)
- Stub HttpService (not HTTParty) in OAW permissions specs to test at
  the correct abstraction layer

Closes #1199

### DIFF
--- a/app/importers/unpaywall_client.rb
+++ b/app/importers/unpaywall_client.rb
@@ -8,7 +8,8 @@ class UnpaywallClient
       json = JSON.parse(HttpService.get(find_url))
     else
       find_url = "https://api.unpaywall.org/v2/search/?query=#{CGI.escape(publication.title)}&email=openaccess@psu.edu"
-      json = JSON.parse(HttpService.get(find_url))['results'].blank? ? '' : JSON.parse(HttpService.get(find_url))['results'].first['response']
+      result = JSON.parse(HttpService.get(find_url))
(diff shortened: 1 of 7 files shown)


jlandiseigsti (approved):
Looks good to me!

PR #1223, opened 2026-04-22
NIH grant metadata import

### LINKED ISSUE #150: Enhancements - Pull from NIH for contract/grants
Pull from the NIH for grants/Contracts
This may be a good place for us to start investigating.
https://exporter.nih.gov/
https://grants.nih.gov/grants/oer.htm

EricDurante:
closes #150


EricDurante:
This is ready for review.

ajkiessl (approved):
This LGTM. I suppose we'll want to keep an eye on the agency names from here. I was expecting just "NIH", but looks like there are others. Our model validations will let us know, though.

EricDurante:
@ajkiessl 

> I suppose we'll want to keep an eye on the agency names from here. I was expecting just "NIH", but looks like there are others. Our model validations will let us know, though.

Yeah, there are a handful of different agencies represented in this dataset, but they are consistently named. I don't know if additional agencies will ever appear in the data, but my intention was that if any ever do, then the importer will throw a validation error, we'll detect that, and then we'll decide how to handle the new agency (probably by adding it to our list). The point is that we always want to make sure that the values that we're ingesting for identifier and agency_name from every source are uniform and expected.

PR #1230, opened 2026-05-06
1225 grant importing changes

### LINKED ISSUE #1225: Remove old Web of Science grant data and stop importing
## Background

Early on in RMD development, we had access to a static dataset from Web of Science that included publication and grant metadata for about a five year period. We built an importer to identify the metadata in that dataset that was relevant to Penn State and to ingest that subset of data into RMD. At the time, RMD only had one other source of metadata about grants which was a set of static files downloaded from the National Science Foundation website. 

In RMD, metadata about individual grants is represented by `Grant` records. Those records have the typical `id` primary key as their identifier in the RMD database, but they also have some other identifier fields that represent the identifiers that were assigned to the grants by the agencies that issued them. There are also fields that are used for storing the name of the issuing agency since identifiers aren't necessarily unique across multiple agencies. Specifically, there are fields called `identifier` and `agency_name`. There are also fields called `wos_identifier` and `wos_agency_name`. The first set of fields was used for storing information that was imported from NSF (and was intended to store information from other sources to be added in the future). The second set of fields was used to store the same information (identifier and agency name) specifically for records that were imported from Web of Science.

## Problem

The reason for having these two separate sets of fields for storing the same type of information is that there was no way to compare the identifiers or agency names that were imported from Web of Science with the identifiers and agency names that were imported from other sources like NSF - the naming/formatting/entry of the data in Web of Science was too wildly messy and inconsistent to be used for anything except for comparing records from new Web of Science imports with records from other, previous Web of Science imports.

The incomparability of the Web of Science data means that we undoubtedly already have a lot of duplicate grant data in RMD (since Web of Science has messy information about NSF grants). It also means that we have no way of knowing the extent of that duplication, and that we don't have any practical way to eliminate it. Now that we're starting to import grant metadata from other primary sources (like NIH), that problem is only going to get worse. Because Web of Science is only a secondhand source of grant information, they also have messy metadata about grants from NIH and other agencies. If we keep the data from Web of Science as we import from additional new sources, then we'll end up with a grant dataset in RMD that is so messy and duplicative that it won't be very useful for anything.

## Proposed Solution

We need to eliminate Web of Science as a source of grant metadata for RMD and stick to primary sources of that data only - like imports directly from NIH and NSF. If we discover another secondhand source of grant information that has clean, consistent metadata, then we may be able to use that as well, but only if it's really reliable.

### Steps to implement

1. Expunging existing Web of Science grant data from RMD should be really easy. All that we need to do is delete every `Grant` record that has non-null/non-blank value in the `wos_identifier` field. There are already cascading deletes set up on the foreign keys in the `grants` table, so associated join records will be properly deleted as well.
2. We should update the existing Web of Science importer so that it no longer imports grants or associated records.
3. We should drop the `wos_identifier` and `wos_agency_name` columns from the `grants` table.
4. Dropping the DB columns is going to break some of our other code, so we should find all references to those columns and update accordingly.
5. Specifically, we should update any other importers that ingest grant metadata so that they no longer have to attempt to deal with the Web of Science messiness.

There are a couple of other small, somewhat related items that we might want to tackle here as well:

1. The current NSF importer hard-codes "National Science Foundation" as the agency name for all imported grant records even though the metadata contains the agency name. We may want clear out those hard-coded values that we recorded, then update the importer to save the value from the provided metadata and reimport those values.
2. I think that it would be a good idea to add an `import_source` column to the `grants` table. Similar to the `Publication` model, the value in this column would be an enum from a list of our sources of grant metadata, i.e. `['NSF', 'NIH']`. We'd then update each grant metadata importer to record the `import_source` when it creates a new record. We would also need to set the `import_source` for existing grant records. After the Web of Science grant records are deleted, then we can assume that all other existing records came from NSF as long as we do this before we release the new NIH grant importer.
3. I think that we should change how we match against grants that we've already imported when we're importing new grant metadata. Currently we match on both the grant identifier and the name of the awarding agency since grant identifiers aren't universally unique. However, I think that in practice it's very unlikely that we'll have identifier collisions between grants from different agencies. Even if we occasionally did, the worst that would happen would be that we might miss importing a grant here and there. There doesn't seem to be much standardization around how grant identifiers are presented in metadata, but there's even less with regard to how agency names are presented. A change or discrepancy in the agency name could lead us to import a lot of duplicate grant metadata. So I think it would be a better compromise if we start matching only on the identifier.

EricDurante:
closes #1225

@ajkiessl @Smullz622 does this look good to you?

There are a few additional small changes that I want to make that are related to this, but they'll need to be made after this work is released to production and after we've done a little bit of production data manipulation.


ajkiessl (approved):
This makes sense and LGTM.  I'll let @Smullz622 also give a review, though, since this changes some of the stuff she just worked on.  I suppose the list: [NSFC, DOE, NASA, NHLBI, NIMH, NIDA, NIA, NICHD] would be a good starting point for more funding imports from primary sources.

EricDurante:
@ajkiessl Yeah, good point. Those agencies seemed to be pretty well represented in the WoS data. Granted, that data wasn't recent.

Smullz622 (approved):
LGTM!

PR #1237, opened 2026-05-14
Add open access scholarly work publish webhook

### LINKED ISSUE #1232: Open Access Scholarly Work publish webhook
RMD needs a webhook endpoint to accept a ScholarSphere URL for a `ScholarsphereWorkDeposit` that is now complete in ScholarSphere.  ScholarSphere will send a payload like:

```
{
  scholarsphere_work_url: "/resources/{work_uuid}"
}
```

RMD should use this URL to identify the `ScholarsphereWorkDeposit` with this URL as it's `draft_scholarsphere_work_deposit_url`, and create an associated ScholarSphere `OpenAccessLocation` with that URL.  If the webhook does not find a `ScholarsphereWorkDeposit`, it can just return `ok`.

This endpoint should authenticate via a webhook token.

karthikreddy-02:
This PR implements a new webhook endpoint that receives notifications from ScholarSphere 
when a scholarly work deposit has been successfully published.

Migration: Adds `draft_scholarsphere_work_deposit_url` column to `scholarsphere_work_deposits`.  This stores the relative URL (e.g. `/resources/{uuid}`) that ScholarSphere returns when a deposit is initially submitted, so it can later be matched against the incoming webhook payload.

Route: Adds `POST /webhooks/open_access_works` alongside the existing `scholarsphere_events` webhook.

Controller: Adds `open_access_works` action to `WebhooksController` that:
  - Authenticates requests via the existing `X-API-KEY` header mechanism
  - Looks up the `ScholarsphereWorkDeposit` record by `draft_scholarsphere_work_deposit_url`
  - If found, calls `record_success` to mark the deposit as successful and create an `OpenAccessLocation`
  - Always returns `200 ok` regardless of whether a matching deposit is found
  
 Specs: Adds full request spec coverage for the new endpoint
  
Closes #1232


app/controllers/webhooks_controller.rb
```diff
@@ -16,6 +16,35 @@ def scholarsphere_events
...
+        # creates/updates the OpenAccessLocation with source SCHOLARSPHERE,
+        # and destroys the associated file uploads.
+        deposit.record_success(full_url)
+      end
+      # If no matching deposit is found, we silently no-op — per the ticket,
+      # a success response is returned either way.
+    end
+
+    render plain: 'ok'
+  end
```
ajkiessl: These comments can be removed.  I usually include comments if they help to understand some complex business logic influencing the code.  These comments just describe the code, which the code does itself.

config/routes.rb
```diff
@@ -124,4 +124,5 @@
   put 'education_history_items/:id' => 'education_history_items#update', as: :education_history_item
 
   post 'webhooks/scholarsphere_events' => 'webhooks#scholarsphere_events', as: :scholarsphere_events_webhook
+  post 'webhooks/open_access_works' => 'webhooks#open_access_works', as: :open_access_works_webhook
```
ajkiessl: I think we need to rethink the naming for these webhook endpoints.  The controller actions will need renamed too.  `scholarsphere_events` doesn't quite make sense for that action anymore because we have two ScholarSphere events now — one that deletes `OpenAccessLocations` and one that creates `OpenAccessLocations`.  So, you could maybe do something more like:

```
namespace :webhooks do
    namespace :scholarsphere do
      post 'work_withdrawn' ...
      post 'open_access_work_published' ...
    end
  end
```

And then keep the existing `scholarsphere_events` endpoint for backwards compatibility until it can be removed.  Add a `TODO` comment above that line to indicate it can be removed once the code in ScholarSphere has been changed to point to the new endpoint.  We should also add a GH issue to remove that endpoint.

ajkiessl (changes requested):
I have a couple requests below 👇

config/routes.rb
ajkiessl: Also, we should add an issue to the scholarsphere repo to point to the new location after this code is deployed.

app/controllers/webhooks_controller.rb
```diff
@@ -16,6 +16,35 @@ def scholarsphere_events
     end
   end
 
+  def open_access_works
+    # The scholarsphere_work_url param contains the relative path e.g. "/resources/{uuid}"
+    # sent by ScholarSphere when a deposit finishes publishing.
+    # If it's missing, we still return 'ok' — ScholarSphere expects a success response regardless.
+    if params[:scholarsphere_work_url].present?
```
ajkiessl: We should also return a `:bad_request` response if this param is not present.

config/routes.rb
```diff
@@ -123,5 +123,14 @@
...
+  namespace :webhooks do
+    scope '/scholarsphere' do
+      post 'work_withdrawn' => 'scholarsphere#work_withdrawn', as: :scholarsphere_work_withdrawn
+      post 'open_access_work_published' => 'scholarsphere#open_access_work_published', as: :scholarsphere_open_access_work_published
+    end
+  end
+
+  # TODO: Remove once ScholarSphere has been updated to use the new namespaced endpoints above.
+  # See: https://github.com/psu-libraries/researcher-metadata/issues/XXXX
   post 'webhooks/scholarsphere_events' => 'webhooks#scholarsphere_events', as: :scholarsphere_events_webhook
```
ajkiessl: I should have clarified: this can point to the new controller action and the old controller can be deleted.

```
post 'webhooks/scholarsphere_events' => 'webhooks/scholarsphere#work_withdrawn', as: :scholarsphere_events_webhook
```

Also, can you add that GH issue and point to it?

db/schema.rb
```diff
@@ -572,6 +572,7 @@
     t.bigint "deputy_user_id"
     t.string "deposit_workflow"
     t.integer "activity_insight_oa_file_id"
+    t.string "draft_scholarsphere_work_deposit_url"
```
ajkiessl, line 575: This needs to be tested in the model `ScholarsphereWorkDeposit` spec.

app/controllers/webhooks/scholarsphere_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+        locations.destroy_all
+        head :no_content
+      else
+        head :bad_request
+      end
+    end
+
+    def open_access_work_published
+      if params[:scholarsphere_work_url].present?
```
ajkiessl: If this is not present it should return a bad request response.

ajkiessl (changes requested):

ajkiessl (approved):
LGTM

PR #1247, opened 2026-05-28
1227 export metadata to schol

### LINKED ISSUE #1227: Open Access upload: Export metadata from RMD to Scholarsphere
Depends on https://github.com/psu-libraries/scholarsphere/issues/1957

Replace the form below (at `profile/publications/{id}/open_access/edit`) with a link:

Image

When this link is clicked, the metadata for this `Publication` should be sent to ScholarSphere via the updated `/ingest` API endpoint (https://github.com/psu-libraries/scholarsphere/issues/1957).

We should still create a `ScholarsphereWorkDeposit` when this happens, even though this will now hold much less metadata than before.  We will need to add some validation contexts to the model, since the lack of metadata will likely cause validation errors.  The `ScholarsphereDepositService` will need modified too:  we don't want an email sent to the user when the deposit happens, and we want the Scholarsphere client to accept `publish: false` when we create a draft.

The response from this endpoint — if successful — will return an `edit_url` that directs to the "Files" upload page in ScholarSphere.  The user should be redirected to this URL.  There will also be a `url`, which directs to the work's show page in ScholarSphere.  An `OpenAccessLocation` can be created with this URL for this `Publication`.

_Something to discuss: this `OpenAccessLocation` is not confirmed to be published, and I'm not sure if it's possible to confirm.  Will this be a problem?_

_Also, what happens when it fails?_

jlandiseigsti:
closes #1227


app/controllers/open_access_publications_controller.rb
```diff
@@ -145,35 +149,24 @@ def scholarsphere_deposit_form
...
-    flash[:notice] = I18n.t('profile.open_access_publications.create_scholarsphere_deposit.success')
-    redirect_to edit_profile_publications_path
+    service = ScholarsphereDepositService.new(@deposit, current_user)
+    edit_url = service.create_draft
+    redirect_to edit_url, allow_other_host: true
   rescue ActiveRecord::RecordInvalid
     @form = OpenAccessURLForm.new
     flash.now[:alert] = @deposit.errors.full_messages.join(', ')
     render :edit
+  rescue ScholarsphereDepositService::DepositFailed
```
ajkiessl: Somewhere in here we should send `record_failure` to the `deposit`.

app/controllers/open_access_publications_controller.rb
```diff
@@ -145,35 +149,24 @@ def scholarsphere_deposit_form
...
       @authorship.update!(updated_by_owner_at: Time.current)
     end
-
-    ScholarsphereUploadJob.perform_later(@deposit.id, current_user.id)
-
-    flash[:notice] = I18n.t('profile.open_access_publications.create_scholarsphere_deposit.success')
-    redirect_to edit_profile_publications_path
+    service = ScholarsphereDepositService.new(@deposit, current_user)
+    edit_url = service.create_draft
+    redirect_to edit_url, allow_other_host: true
```
ajkiessl, line 161: Either somewhere here, or in the happy path in `ScholarsphereDepositService#create_draft` logic, we should store the draft work's url in `draft_scholarsphere_work_deposit_url`.  This will be used later in the webhook from ScholarSphere to update the deposit as successful.  It will also be displayed to admins so they can check the deposit from RMD.  The `draft_scholarsphere_work_deposit_url` is not an attribute added yet, but it should be an attribute for `ScholarsphereWorkDeposit`.  This might conflict with this a bit: https://github.com/psu-libraries/researcher-metadata/pull/1237/changes#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R575 but it should be a minor conflict.

ajkiessl (changes requested):
This is looking good, and almost there.  I have a few requests below 👇

app/services/scholarsphere_deposit_service.rb
```diff
@@ -8,6 +8,34 @@ def initialize(deposit, user)
...
+
+    response = ingest.create
+    raise DepositFailed.new('Something went wrong.') unless response
+
+    response_body = JSON.parse(response.body)
+
+    # Draft works will return a 201
+    if response.status == 201 || !response_body['edit_url'].nil?
+      edit_url = "#{ResearcherMetadata::Application.scholarsphere_base_uri}#{response_body['edit_url']}"
+      deposit.update(draft_scholarsphere_work_deposit_url: edit_url)
```
ajkiessl: Sorry, I should have clarified.  The `draft_scholarsphere_work_deposit_url` should be the 'url' in the response body not the 'edit_url'.  The `url` is associated with the `Work` in ScholarSphere and is the url that will be used in the webhook to update this as successful when complete.

ajkiessl (changes requested):

app/services/scholarsphere_deposit_service.rb
jlandiseigsti: Ah, thanks for clarifying. I'll make the change!

jlandiseigsti: Is it looking for the full url or just the path?

jlandiseigsti: Actually, I assume it's full so admins can easy go to it.

ajkiessl (approved):
LGTM

PR #1255, opened 2026-06-05
Remove vendor/cache and its contents

### LINKED ISSUE #1129: Consider not caching gems
We cache our gems in `vendor/cache` and store them with the codebase in source control.  We originally did this to prevent the downloading of gems on deploy when we were using chef for infrastructure and capistrano for deployments.  Now that we build images with docker in CI and just run the images in production (which will already have the gems cached), I'm not sure this is needed anymore.

ajkiessl:
Using vendor/cache is redundant since our images install the gems on build.  This PR removes vendor/cache and it's contents.  It also removes the Docker build steps that include vendor/cache.

closes #1129 

I decided to just fix this now, since there are mismatches with my local env, and I kept having to clean up the git history.


bryanlopezrr (approved):
lgtm

PR #1256, opened 2026-06-05
Adds Scholarsphere OA Scholarly Work metadata fields to ScholarsphereWorkDeposit#metadata

### LINKED ISSUE #1253: Set `open_access` boolean when exporting to ScholarSphere via Standard OA Workflow
This should be conditional on DOI.  If we don't have a DOI, there's no need to check for permissions.  _So_, if there is no DOI we don't set the `open_access` boolean on `WorkVersion`, and if there is a DOI we do set the `open_access` boolean on `WorkVersion`.

ajkiessl:
This metadata is dependent on the OA workflow pathway and DOI.

closes #1253 

Scholarsphere needed some changes too: https://github.com/psu-libraries/scholarsphere/pull/1997


vendor/cache/scholarsphere-client-0.5.3.gem
jlandiseigsti, line 1: Did you mean to remove this cached gem?

ajkiessl: Yes, I completely removed the vendor/cache here: https://github.com/psu-libraries/researcher-metadata/pull/1255

But I was simultaneously working on this branch, so I think this one didn't get removed, or something.

jlandiseigsti (approved):
Looks good to me!

PR #1261, opened 2026-06-05
Alert user when ScholarSphere upload fails

### LINKED ISSUE #1248: Turn off ScholarSphere Standard OA Workflow export if failed
If the export to ScholarSphere fails when clicking the link to start the upload in ScholarSphere, the link should no longer display and there should be some help text to let the user know it failed and they should contact administrators.

ajkiessl:
Adds logic to hide "Deposit to ScholarSphere" button when upload has already failed.  Show an alert message instead.  Adds tests for the view component and refactors text into translation file.

closes #1248 

Screenshot 2026-06-05 at 3 18 20 PM


jlandiseigsti (dismissed):
Looks good to me

PR #1266, opened 2026-06-09
1241 a11y fixes

### LINKED ISSUE #1241: Accessibility Fixes
[Research Metadata Database Testing Results.docx](https://github.com/user-attachments/files/28069746/Research.Metadata.Database.Testing.Results.docx)

jlandiseigsti:
closes #1241
Includes proper heading use, contrast and spacing issue, html language, and use of aria landmarks.

### DIFF
--- a/app/views/public/home.html.erb
+++ b/app/views/public/home.html.erb
@@ -1,5 +1,5 @@
 <div id="home-page">
-  <div class="top-section">
+  <section class="top-section">
     <div class="container">
       <div class="row">
         <div class="col-lg-8">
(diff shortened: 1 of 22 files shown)


app/views/public/home.html.erb
```diff
@@ -13,29 +13,29 @@
...
     <div class="row text-center">
       <div class="col-lg-4">
         <%= image_tag('metadata.png', alt: 'Researcher Metadata') %>
-        <h4>Researcher Metadata</h4>
+        <h3 class="h4">Researcher Metadata</h3>
         <p>
           We maintain an aggregation of publications, presentations, performances, advising, news articles, and bios for Penn State faculty from several major sources including Pure and Activity Insight.
         </p>
-        <a class="btn btn-primary" href="https://github.com/psu-libraries/researcher-metadata/wiki">Learn More</a>
+        <a class="btn btn-primary" href="https://github.com/psu-libraries/researcher-metadata/wiki">Learn More about Researcher Metadata</a>
```
ajkiessl: I know there's not much context in the accessibility doc about this, but if the reason for this change is:

"This needs to be 'Learn More about Researcher Metadata' because 'Learn More' gives limited context for users using screen readers" then I think making "about Researcher Metadata" hidden text is preferable here.

If the reason is:

"This needs to be 'Learn More about Researcher Metadata' because it is generally confusing and limits context even for visual users", then I would agree with the change.  However, I don't agree with that premise, the context is quite clear for visual users.

So, I think the "about Researcher Metadata" should be hidden text to help with screen readers while preventing this button from looking clunky and standing out too much from the rest of the screen.

ajkiessl (approved):
LGTM

PR #1269, opened 2026-06-10
Fix flaky file version review spec

jvitorbarros15:
## Summary
- Add `.order(:id)` to `Publication.file_version_check_failed` so the admin File Version Review page lists publications in a stable order.
- Make `file_version_review_spec.rb` click the edit link by its unique href instead of picking a table row by position (`find_all('tr')[1]`), resolving the old TODO about flakiness in that spec.

## Why
CI failed on `file_version_review_spec.rb:48` on an unrelated branch: the scope has no `ORDER BY`, so Postgres returns publications in arbitrary heap order, and the test assumed the first table row belonged to `pub1`. Any branch that shifts the test database's row layout can trip this flake.

The scope's other two usages are unaffected by the ordering: a `.count` on the OA dashboard (Rails strips `ORDER BY` from count queries) and an order-independent `contain_exactly` assertion in `publication_spec.rb`.

## Testing
- `spec/integration/admin/activity_insight_oa_workflow/file_version_review_spec.rb` — 3 examples, 0 failures
- `spec/component/models/publication_spec.rb` — 456 examples, 0 failures
- `spec/component/components/activity_insight_oa_dashboard_component_spec.rb` — 19 examples, 0 failures
- RuboCop: no offenses

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -232,6 +232,7 @@ class Publication < ApplicationRecord
       .where(%{EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version = 'unknown')})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND publications.preferred_version = activity_insight_oa_files.version)})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version IS NULL)})
+      .order(:id)
   }
   scope :wrong_file_version_base, -> {
     nonflagged_activity_insight_oa_publication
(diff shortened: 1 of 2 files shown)


app/models/publication.rb
```diff
@@ -232,6 +232,7 @@ def self.preferred_version_options
       .where(%{EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version = 'unknown')})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND publications.preferred_version = activity_insight_oa_files.version)})
       .where(%{NOT EXISTS (SELECT * FROM activity_insight_oa_files WHERE activity_insight_oa_files.publication_id = publications.id AND activity_insight_oa_files.version IS NULL)})
+      .order(:id)
```
ajkiessl, line 235: I'm not certain we want this here.  If the flaky test can be fixed with your changes to the actual test, then that should be sufficient.  If we _need_ to order, then we should do that closer to the display level, since that's what the ordering is for.

jvitorbarros15: Hey Alex, I agree. I just fixed it in my tests too and it passes the tests. I will remove this PR then.

PR #1270, opened 2026-06-10
Rename open_access to open_access_upload in scholarsphere metadata export

karthikreddy-02:
Depends on: psu-libraries/scholarsphere#1996

Renames `open_access` to `open_access_upload` in `ScholarsphereWorkDeposit#metadata` to match the renamed column in ScholarSphere.


ajkiessl (approved):
LGTM

Issue #1243, opened 2026-05-26
Import Google Scholar citation


jvitorbarros15:
Use Google Scholar scraper to pull citation information from users and individual publication and H-index.

- When a faculty member already has a Google Scholar ID, use it to pull their total citations and H-index.
- If faculty does not have a Google Scholar ID, name match them in Google Scholar. Use DOI from one or more of their associated publications to confirm the correct Google Scholar profile. 
- If DOI match is confirmed, import Google Scholar ID, total citations, and H-index.
- If no DOI match is found, try matching by publication title as a fallback.
- Use a separate importer for publication-level citations.
- Implement as Cron task

PR #1272, opened 2026-06-11
Add/google scholar integration

jvitorbarros15:
Partially Closes issue #1243: 

  **Profile discovery (`GoogleScholarProfileImporter`)**
  - Resolves a user's Scholar ID from `google_scholar_id` if
  already set, or extracts it from the Activity Insight
  `ai_google_scholar` URL
  - Falls back to name-based discovery via ScraperAPI Google
  search, fetching each candidate profile and verifying it
  with `GoogleScholarProfileMatcher`
  - Matcher rejects candidates whose profile
  institution/email domain is not Penn State, then confirms
  identity by DOI overlap (≥2 shared DOIs, or ≥1 if
  PSU-affiliated) or fuzzy title match (≥2 titles with ≥0.90
  trigram similarity, publication year within ±2)
  - Skips users with 0 publications (nothing to match
  against) to conserve ScraperAPI credits, and stops the run
  if the credit budget is exhausted
  - Tracks `google_scholar_checked_at` /
  `google_scholar_not_found` on users so re-runs only retry
  stale records (default 120-day refresh) and don't re-search
  known misses

  **Metrics import**
  - Stores Google Scholar h-index and total citation count on the user from
  the matched profile
  - `GoogleScholarPublicationCitationImporter` fetches
  per-publication citation counts by DOI

  **Admin**
  - Shows Scholar ID (linked to the profile), citation
  metrics, and fetch-tracking fields on the user views


app/models/user.rb
```diff
@@ -56,6 +56,15 @@ class OmniauthError < RuntimeError; end
...
   scope :active, -> { where.not(psu_identity: nil).where("psu_identity->'data'->>'affiliation' != '[\"MEMBER\"]'") }
 
+  scope :admin_search, lambda { |query|
+    return all if query.blank?
+
+    query.to_s.split.map { |t| "%#{t}%" }.reduce(all) do |s, term|
+      s.where('first_name ILIKE ? OR middle_name ILIKE ? OR last_name ILIKE ? OR webaccess_id ILIKE ?',
+              term, term, term, term)
+    end
+  }
```
ajkiessl: What happens if a user searches by Activity Insight ID or ORCiD?  I'm not sure we want to make this change.  I have not had a product owner suggest that they would like this.  I think it would be better to remove this and open an issue and wait to see if it's something our PO/admin users would like.

docs/superpowers/specs/2026-06-08-skip-zero-pub-users-in-discovery-design.md
```diff
@@ -0,0 +1,49 @@
...
+
+Add one spec case to the `GoogleScholarProfileImporter` spec:
+
+- When a user has 0 publications and no `google_scholar_id` / `ai_google_scholar` URL, `scraper.search_profiles` must not be called and the user record must not be updated.
+
+Existing specs for paths 1 and 2 (known ID) need no changes.
+
+## Credits saved
+
+Per skipped user: 25 credits (search call avoided). For a faculty roster with many researchers who have no publication record in the local DB, this compounds across every import run.
```
ajkiessl: Can this be removed?

app/importers/google_scholar_profile_importer.rb
```diff
@@ -0,0 +1,117 @@
...
+        break
+      end
+
+      import_user(user)
+      pbar.increment
+    end
+
+    pbar.finish
+
+    sleep 30
```
ajkiessl: Why `sleep 30` here?

ajkiessl (commented):
There's a lot here, so I'm going to leave some comments in chunks as I go through.  I still have a lot left to review.

app/importers/google_scholar_profile_importer.rb
jvitorbarros15: This was from my previous scraperapi code, the reason was because I was fetching the credits from the scraperAPI platform and they had a little delay to update it. But we dont need it anymore, just removed it

app/models/user.rb
jvitorbarros15: That was because the user search wasn't working properly from the admin view searching a full name like "Zhou Sha" returned nothing. I was running into this constantly while verifying Scholar imports by name, so I added a scope that splits the query into terms and matches each one against first/middle/last name and webaccess ID. But I agree, it replaces the default search and drops fields it used to cover (like ORCiD and Activity Insight ID). I'll revert it, I guess is not that necessary, but it does help a lot finding people

docs/superpowers/specs/2026-06-08-skip-zero-pub-users-in-discovery-design.md
jvitorbarros15: Yes, sorry. Pushed it by accident.

jvitorbarros15:
## Code Review: 6 Confirmed Issues

### Critical Bugs

**A. Transient network error → permanent user exclusion** (app/importers/google_scholar_profile_importer.rb:60-75)
- discover_profile returns nil indistinguishably for "no results" and "ScraperAPI timeout"
- Sets google_scholar_not_found: true either way
- User then filtered out of all future runs until admin resets flag
- Fix: Signal network errors distinctly; only set not_found: true on clean no-results

**B. PublicationFuzzyMatcher dead code + broken Rails 7.2 API** (lib/utilities/publication_fuzzy_matcher.rb)
- Zero callers — never used
- connection.execute($1, $2) doesn't bind params; 2nd arg is log name, not binds
- If called: $1/$2 never bound → PG error → rescues to 0.0 → silent match failure
- Fix: Delete the file (logic already in GoogleScholarProfileMatcher)

### Performance Issues

**C. N+1 in title_match_count** (app/models/google_scholar_profile_matcher.rb:82-90)
- 1 DB query per Scholar publication
- Scholar profiles have 100+ pubs; 3 candidates × 200 pubs = 600+ sequential queries per user
- doi_match_count solves the same problem in 1 query

**D. Budget cap doubled by split scraper instances** (app/importers/google_scholar_importer.rb)
- Profile importer + citation importer each create own GoogleScholarScraper instance
- Combined run can spend ~2N with CREDIT_BUDGET=N
- Secondary: citation importer loop never checks credit_budget_exceeded? in its iteration

**F. Full profile fetch per candidate before filtering** (app/importers/google_scholar_profile_importer.rb:77-88)
- Fetches entire paginated profile (100+ pubs + sleeps) for every search candidate before affiliation check
- Structured search already has title/snippet to cheap-filter

### Code Quality

**E. Duplicated URL parsing** (scholar_id_from_ai_url vs scholar_id_from_href)
- Same logic in two places; bug fix in one won't propagate

**G. Minor: boolean style inconsistency** (app/importers/google_scholar_profile_importer.rb:44)
- google_scholar_not_found = FALSE (works) but codebase uses IS FALSE

app/importers/google_scholar_profile_importer.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      if candidates.nil?
+        Rails.logger.warn("GoogleScholarProfileImporter: skipping user #{user.id} — ScraperAPI error during search")
+        return :scraper_error
+      end
+
+      candidates.each do |candidate|
+        profile = scraper.fetch_profile(candidate[:scholar_id])
+        next unless profile
+
```
jvitorbarros15, line 89: If `fetch_profile` returns `nil` here due to a transient ScraperAPI error (timeout, 403, etc.), this candidate is silently skipped. If no other candidates succeed, `discover_profile` returns `nil`. Back in `import_by_discovery`, `nil != :scraper_error` is `true`, so the user gets flagged `google_scholar_not_found: true` — even though we never actually determined there was no match. That user is then excluded from discovery for the full `refresh_days` window (default 120 days) on what could have been a one-off network blip.

app/importers/google_scholar_profile_importer.rb
```diff
@@ -0,0 +1,124 @@
...
+                 matcher_class: GoogleScholarProfileMatcher,
+                 refresh_days: ENV.fetch('GOOGLE_SCHOLAR_REFRESH_DAYS', '120').to_i)
+    @scraper = scraper
+    @matcher_class = matcher_class
+    @refresh_days = refresh_days
+  end
+
+  def call
+    users = importable_users
+    users = users.limit(ENV['LIMIT'].to_i) if ENV['LIMIT'].present?
```
ajkiessl: I would either remove the `LIMIT` environment variable and make it a constant or make the environment variable more descriptive.

spec/component/importers/google_scholar_profile_importer_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+        expect(scraper).not_to have_received(:fetch_profile)
+      end
+    end
+
+    context 'when a user is flagged not found but now has a Scholar ID' do
+      let!(:user) do
+        create(:user, :with_psu_identity,
+               google_scholar_id: 'late-added-id',
+               google_scholar_not_found: true)
+      end
```
ajkiessl, line 347: I couldn't figure this out from the code, and maybe I'm not understanding correctly, but how would someone end up with a Scholar ID if they've already been flagged as "not found".  Since it looks like they are getting filtered out of the profile fetching process.  Would it need to be added manually?

jvitorbarros15: Yes, when they get added it later on manually. So after the faculty gives us their scholar ID, we will run the scraper on them even if it was not found before

app/importers/google_scholar_profile_importer.rb
jvitorbarros15: Removed!

ajkiessl (dismissed):
This looks good to me.  It's well tested, well documented, and refactored.  The specs cover all our acceptance criteria.  The only thing blocking this looks like a merge conflict with the db/schema.  Once that's fixed this can be merged to `main`.

A couple things to consider for small future improvements:

- Add some tests in the model specs for the new db fields
- Add `google_scholar_imported_at` to the admin show display
- Make sure the API returns these new fields in the UserProfile and Publication endpoints
- Make sure the API is documented with these new fields

These improvements can be added to a separate GH issue.

ajkiessl (approved):

PR #1278, opened 2026-06-22
Rescue and store errors when PubMed fetch fails

ajkiessl:
When running the NIH import in ULDEV, it repeatedly fails after about 1.5 hours with a TCP connection error.  I'm not too sure the cause, but since it's a connection error, and right now appears to maybe just be one record, I think it makes sense to rescue it, log it, and continue the import.


Smullz622 (approved):
LGTM

PR #1288, opened 2026-06-29
Fix Google Scholar Publication Citation Import

ajkiessl:
The per-publication citation importer was not importing anything.  There was an issue with the scraper query and no DOI is returned in the Google Scholar output, so we can't match on DOI.  We _do_ search by DOI, but now we use a fuzzy-title match to confirm.


jlandiseigsti (approved):
LGTM

PR #1290, opened 2026-07-13
Delete ScholarsphereWorkDeposits when withdrawn and use full URLs

ajkiessl:
For more details check out https://github.com/psu-libraries/scholarsphere/pull/2039 .

This PR adds functionality to the "withdrawn" webhook to delete `ScholarsphereWorkDeposits` related to the withdrawn work.  Also, changes to how we store the draft URL on creation so that the full url is stored.  Both the "withdrawn" and "published" webhooks are using full urls now.


jlandiseigsti (approved):
Looks good to me

PR #1292, opened 2026-07-15
Hotfix: Rescue StandardError at top-level year iteration in NSF import

ajkiessl:
This is mostly to catch network errors when grabbing the nsf data, but will help to catch other things as well.


jlandiseigsti (approved):
LGTM

